PR #1, opened 2017-10-31
fix for hot reloading

revgum:


docker-compose.yml
```diff
@@ -118,6 +118,7 @@ services:
       - RAILS_CACHE_STORE_URL=memcache
     volumes:
       - app:/data/tmp/uploads
+      - application:/data/app
```
jechols: Change this to `"- .:/data"`

PR #11, opened 2018-04-06
adds notes about Docker Stack

revgum:
updates OD2 to Hyrax2.1, adds docker-stack and some notes on how to use it. A generic_work for show'n'tell testing.

Next step would be extending the docker-stack docker-compose.yml to include the database and redis. TODO: @revgum


jechols:
I'm not opposed to making it easier to use Docker, but this docker-stack thing seems a little crazy.  We're basically saying that in order for everybody to be using the same docker setup, we have to have the right versions of all the prerequisites installed on our local system.  Which is one of the reasons to use docker in the first place: we don't need to have Ruby, Rake, Bundler, all the gems, ...

Can you split up your PR?  We have three relatively large changes all in a single PR.  I would almost certainly approve updating to hyrax without a second thought, and I'd like to look at what the generic_work really looks like on its own.  Then we could debate docker-stack in a separate PR without slowing down these other two relatively important pieces.

The commit messages also don't really define what's been changed - the generic work and OD2 updates should really be in their own commits (if nothing else, it would be a lot easier to pull these into their own PR).  It's not clear just glancing at the history that anything other than Docker Stack was implemented here.

jechols:
Actually the more I look into docker-stack, the more I think we should avoid this.  It's reducing the portability of the docker solution, it's adding a bunch of extra dependencies, extra points of failure, a potential maintenance burden, more mental overhead, a new "language" to learn for those of us who already use docker as-is, and a wonderful way to avoid understanding how docker actually works.

That last point is important.  When something "weird" happens in a container, people need to understand how to debug that, which means people are going to have to have a somewhat intimate understanding of how docker works.

And all it really does is keep us from typing `docker-compose up -d`, `docker-compose down`, `docker-compose logs`, etc.  I really don't understand the value here.

If this thing can be configured to be used by individual devs without checking it into the repo, I sure won't try to stop you, but my gut tells me this would be a bad thing to force on the repository.

revgum:
@jechols This is docker, the gem, with some helpers on to facilitate the starting/stoping/cleaning containers. I've been using it with ScholarsArchive development and it's been helping my development workflow. While I agree that it takes us a step away from all having to know how to manage the docker containers, I think the overhead is worthwhile. I'd argue that it's better for development and test (which is all this would be used for) that we don't have to try and figure out what "weird" things happens to a container, kill it, clean it, start it up again.. If that "weird" thing happens again, it's more than likely the fault of the code not docker. 

The nice thing is that this is fully optional.. you can ignore it and run the docker-compose yourself no big deal.

jechols:
It really isn't optional if it's in the repo.  It creates its own docker-compose files and hides those from view, and if I were to ignore those files, my dev environment is suddenly out of sync with yours.  Using standard, well-known docker techniques, such as having docker-compose.yml in the root and adding docker-compose.override.yml for local customizations, becomes a burden instead of just standard usage.

- This requires Ruby, bundler, and installing / compiling all the app's gems locally - which is what docker helps us avoid.
- It adds a layer that wraps docker such that we have to learn new commands instead of what docker-compose already does.
- It adds a point of failure and potential for debugging things that aren't related to our app or docker.
- It doesn't even follow good docker practices, like using a single base compose file, and overriding services / values as necessary for testing / staging / etc.  In the docker-stack approach, if we change the development setup, we have to remember to also change the test setup.
- It wraps docker shell commands in two-gems-distant tasks, making it a pain to trace precisely what a given rake task is actually executing.
- This is a completely Rails-specific approach and what happens here will never extend to general use.  Meaning it's extra mental overhead for those who work on multiple types of projects.

There is nothing this setup provides us that can't be very easily done with docker-compose by itself.  I see a lot of cons and not a single pro.  The absolute worst-case of using docker-compose is having a shell script that sets up aliases for you:

```bash
for cmd in rails rspec bundle rake; do
  alias "$cmd"="docker-compose exec -u $UID workers $cmd"
done
```

That's obviously not a comprehensive list of every alias one might need, but you get the idea.

Most of us on the UO side have already learned how to use docker, and this forces us to either learn a new approach that's completely OD2-specific or else diverge from having a consistent dev environment.

---

Aside from this, the other changes really need to get in their own PRs so we can merge them.  Or at least look at them by themselves.  At the moment I really don't know what is a part of the hyrax 2.1 update, what's part of the generic work, and what's specific to docker-stack.

jechols:
Hmm, I must be misunderstanding something.  To me it doesn't seem like it would be optional if we wanted to keep all our development environments consistent.  Maybe this is something to discuss at our next in-person meeting, because I feel there must be some missing context or something.

Looking forward to your other PRs!

revgum:
@jechols it's optional because you could run `docker-compose` in the same way you always do.. and debug the app, watch logs, run test suite, etc... The existence of `.docker-stack` sub-directory doesn't interfere with that process, and your rails app could run in a container with its mounted volume, etc. 

The `docker-stack` approach is just some syntactic sugar on top of what you already do.. and those docker-compose files live within the `.docker-stack` sub-directory. The dev workflow that I use is to run redis, fedora, and solr as docker containers and the app locally (with its configurations pointing at the services hosted in docker).. I like this because it gives me a better debugging experience (I used remote debugging and breakpoints and all that in vscode).. 

At any rate.. we can chat more about it if you'd like, and maybe you have a better dev workflow that you can describe/show me??

jechols:
Ahhhhhhhh, okay.  I was assuming the docker experience was for the full stack (and couldn't understand why the docker files provided by docker-stack didn't define things like the Rails app itself).

So that's the core difference.  I run the whole thing inside docker, not just the secondary / tertiary services.  What you're proposing is having a setup where you are running some of the stack locally and some inside containers.  My big resistance to that approach is the pain we had with getting local versions of ffmpeg, vips, etc.  It got really difficult for people to test locally in various situations.

I think it would be best to get everybody in a state that could ensure some consistency from dev through staging and into production.  That said, though, I can see how debugging may be easier if Rails is running locally.  Maybe there are cool ways to debug through a container, I have no idea....

I'd still prefer keeping docker-stack out of the repo, but I have to think on this more.

As for a better wokflow, probably not.  I'm all-console.  Vim, tmux, etc.  It helps me jump into it from anywhere if I can do all the work I need to via a terminal and "ssh".  So I'm not liable to have any thoughts if you're more on the IDE side of things :)

revgum:
I'm testing out exposing port 1234 and having the rdebug running in the app container to get the capability to mount to the process and debug inside the app in the container.. it looks quite promising. Then we'll have a consistent docker experience for OD2.

jechols:
That would be really slick!  Though as I say, I'm not as opposed now as I was before; I didn't understand that your dev approach involved an IDE, so I can't push as hard against something when the loss of said thing affects you but not me.

revgum:
Hrm.. yea, I'd really like to get this working alongside the docker-compose you've got. I can get the ruby debugger to attach to the process but it doesn't actually work. This just about does it... but not quite.. I'm going to see if it works with RubyMine and maybe just VSCode isn't playing nice. Dang... close, so close.

```
# Example for exposing the load balancer for local dev; just copy this to docker-compose.override.yml
version: '2'

services:
  lb:
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      # You must uncomment this line if and only if you are running docker-machine
      - $DOCKER_CERT_PATH:$DOCKER_CERT_PATH
    expose:
      - 80
    ports:
      - 80:80
  web:
    command: bash -c "rm -f tmp/pids/server.pid && bundle exec rdebug-ide --host 0.0.0.0 --port 1234 -- bin/rails server -p 3000 -b 0.0.0.0"
    ports:
      - 3000:3000
      - 1234:1234
      - 26162:26162
    expose:
      - 3000
      - 1234
      - 26162
    networks:
      internal:
      external:
```

PR #20, opened 2018-09-06
[no-merge] Valykrized

revgum:
Just trying to hack a Hyrax instance to render a form for a valkyriezed model


revgum:
invalid, just hacking

PR #68, opened 2018-11-09
Establishes a test and development workflow for local and CI 

revgum:


jechols:
I approve this PR

wickr:
Hit a warning a few times, "Warning: apt-key output should not be parsed (stdout is not a terminal)"
Setting `ENV APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=1` in the Dockerfile quieted that down.
I got as far as visiting localhost:3000 to set up a user account, but I can't get anything to load. Not sure if it's something with my local config and port 3000.

revgum:
@wickr, yea I checked that error as well and opted to keep the warning(s) visible in case we noticed more/changing issues with that. *shrug* For localhost:3000 did you do this: https://github.com/OregonDigital/OD2/pull/68/files#diff-04c6e90faac2675aa89e2176d2eec7d8R34

luisgreg99:
I followed the instructions and it worked nicely! I like that we can run a test server along side the dev. I tried it and it worked well when running `rspec`. I only had to copy `docker-compose.override.yml` from the example again like @revgum suggested in the readme. It looks like the example is now using version 3 in this branch.

PR #115, opened 2018-11-16
[WIP] Feature/generic work configuration

straleyb:
This is my first pass at fixes for this PR. Still a bit of work to be done. Metadata needs a large overhaul still as some of the predicates are improper. Also need to figure out the terms thing. Other thing I noticed was basic metadata is included twice. Also still figuring out the linked indexing service thing.


app/models/generic.rb
```diff
@@ -2,19 +2,17 @@
...
+
   # Change this to restrict which works can be added as a child.
   # self.valid_child_concerns = []
   validates :title, presence: { message: 'Your work must have a title.' }
 
-  # oEmbed property
-  property :oembed_url, predicate: ::RDF::Vocab::DC.abstract do |index|
-    index.as :stored_searchable
+  def set_defaults
+    super
```
luisgreg99: @straleyb I tried removing `super` locally since I was getting `super: no superclass method `set_defaults' for #` on my local. Are you getting this error too? I'm thinking maybe we need something like default in ScholarsArchive https://github.com/osulp/Scholars-Archive/blob/master/app/models/default.rb#L22-L23?

straleyb: You know, I didnt even notice that, but yeah, I thought this method was coming from somewhere, so thats why I called super. Thanks for the tip!

straleyb:
Closing for weird cci issues

PR #121, opened 2018-11-19
[No-MERGE] initial commit for development k8s

### LINKED ISSUE #94: Look into kubernetes for local development
### Descriptive summary

Development workflow is currently all docker-based.  Production is kubernetes.  This separation means two things:

- Changes to services requires two separate configuration changes, which could get out of sync
- Developers who need to do ops work in a pinch may not know how to do them since k8s isn't something most of us know

If we can reuse enough of the production k8s config for local dev, we could start easing devs into that flow.

- We need to identify what bits of config need to be created that are development-specific
- We should come up with simple scripts (or at least README docs) for starting and stopping the stack

### Expected behavior

Devs can spin up the stack with kubernetes commands

revgum:
fixes #94 

TODO:
- [x] extend `validate_migrated.sh` to check to see if the appropriate solr collection/core has been created, and create it if not. See: https://github.com/OregonDigital/OD2/blob/master/.circleci/config.yml#L48-L51

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,10 +6,11 @@ tmp/*
 public/branding/*
 node_modules/*
 coverage/*
+k8s/*
+
 Dockerfile
 README.rdoc
 
 fits.log
 .byebug_history
 docker-compose.override.yml
-
--- a/k8s/deployments/deployment-db.yml
+++ b/k8s/deployments/deployment-db.yml
@@ -2,8 +2,7 @@ apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
   annotations:
-    kompose.cmd: kompose convert -o k8s/kompose.yml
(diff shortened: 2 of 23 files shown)


revgum:
So here's where I've left off with kubernetes as a development environment;

* The development k8s will not really match the production/staging 100% 
* There would need to be a way to start each of these (or additional pods) in `test` environment for running tests
* Shutting down kubernetes (to free up ports) requires 
  * `kubectl drain <NODE>`
  * `kubectl delete -f k8s/deployments/<EACH LoadBalancer>.yml`

I'm not sure I'm really sold on the idea yet.. Using Docker compose with local overrides (for port mapping) seems to have less friction and just works.

revgum:
This is sorely out of date and not really needed.

PR #129, opened 2018-11-20
super basic generic metadata config

straleyb:


app/models/solr_document.rb
```diff
@@ -1,4 +1,4 @@
-# frozen_string_literal: true
+# frozen_string_literal true
```
revgum: missing colon that was stripped out inadvertently

spec/forms/hyrax/generic_form_spec.rb
```diff
@@ -3,7 +3,163 @@
...
+  end
+  let(:ability) { double('Ability') }
+
+  before do
+    allow(new_form).to receive(:current_ability).and_return(ability)
+    allow(ability).to receive(:current_user).and_return(user)
+  end
+
+  it 'responds to terms with the proper list of terms' do
+    %i[alternative
```
revgum: Should be able to replace this to use the mapped PROPERTIES

spec/models/generic_spec.rb
```diff
@@ -1,9 +1,20 @@
 # Generated via
-#  `rails generate hyrax:work Generic`
+#  `rails generate hyrax:work Audio`
```
revgum: invalid comment, could remove these

spec/models/generic_spec.rb
```diff
@@ -1,9 +1,20 @@
...
+
+  it 'has a title' do
+    subject.title = ['foo']
+    expect(subject.title).to eq ['foo']
+  end
+
+  describe "metadata" do
+    it "has descriptive generic metadata" do
+      props.each do |prop|
+        expect(subject).to respond_to(:contributor)
```
revgum: not checking the property

spec/presenters/hyrax/generic_presenter_spec.rb
```diff
@@ -3,7 +3,23 @@
...
+          title: ['blah'],
+          depositor: user.user_key,
+          label: "filename.tif")
+  end
+  let(:props) {OregonDigital::GenericMetadata::PROPERTIES.map(&:to_sym)}
+  let(:user) { double(user_key: 'generic person 1')}
+
+  it "delegates the method to solr document" do
+    props.each do |prop|
+      expect(presenter).to delegate_method(prop.to_sym).to(:solr_document)
```
revgum: prop is already a symbol

spec/system/create_generic_spec.rb
```diff
@@ -38,7 +38,6 @@
       fill_in('Creator', with: 'Doe, Jane')
       fill_in('Keyword', with: 'testing')
```
revgum: why was the select removed?

straleyb: Because it was appearing under the fold. We havent made decisions what we want placed where on the form, so this test is less valid. Mostly just testing you can create a generic object with some data and it creates. So I think whats filled out is less important

Issue #30, opened 2018-09-24
All Information Based on Derivative Generation Here


straleyb:
### Descriptive summary

This ticket is a nexus for information about the derivative generation process. Fill out as descriptive of information as possible about the decisions made about how this will be working. It can be an epic ticket that we can breakdown in the future.

### Steps to reproduce the behavior

1. Do this
2. Then do this...

### Related work

Link to related tickets or prior related work here.

jechols:
@vantuyls I am not sure what this ticket would need from me at this point.  We've communicated the pros and cons of various options in Google docs and in meetings, but I'm not sure it would be a good use of anybody's time to be copying huge documents into github issues.

I think what I'm getting at here is this: what is the goal of this ticket?  What specific information is being requested?  I can certainly take an hour or two to add everything I've discussed or researched about derivatives, but I don't believe that will translate into actionable items, and I feel it would be a waste of my time.

Are we looking for programming tasks to be added here?  If so, we don't have any yet.  Are we looking to present our findings to stakeholders?  If so, I think meetings and the existing documentation are going to serve us better.  Are we looking for a place to put all documentation links?  If so, I think the wiki would be a better choice, because issues eventually get closed or lost in the clutter.

revgum:
@vantuyls @jechols Probably this document, the cost analysis for derivative generation https://docs.google.com/document/d/1NBJ-JhPGjTK2hzYGwepRuXz_QMofTx6h6bbazuIvFwI/edit

straleyb:
@jechols I think this would be a good place to start getting info together for actionable items. 

Can we include a step by step guide on how this will all connect to SA? 
What would be a good way to pull them from the RAIS server into the application? 
How have you done it previously?
What would those URLs look like? 

Id like this ticket to be a place where we can start building out tickets with actionable items to get this functionality into SA, wether it is server related setup and configuration, or integration into SA. That would be extremely useful here.

jechols:
Lemme try to answer some of that.

As I don't know SA pretty much at all, I can't directly speak to it, but RAIS is a general-use IIIF Image server.  As long as you plug in a IIIF-compliant client (like OpenSeadragon), it "just works".

The magic is all in the IIIF Image manifest files, e.g., https://oregonnews.uoregon.edu/images/iiif/batch_hoodrivercountylibrarydistrict_golden_ver01%2Fdata%2F2003260222%2F00000000503%2F1899110801%2F0373.jp2/info.json

[...]

PR #141, opened 2018-11-24
adds kakadu, rais (iiif), a service to make images work with UV and rais

### LINKED ISSUE #54: IIIF Configuration
### Descriptive summary

RIIIF Needs to be configured early on to verify that derivatives are being processed properly. 

https://github.com/curationexperts/riiif

A Ruby IIIF image server as a rails engine

This is a good solution early on, while the systems team configures and sets up the servers and configurations needed to get this working with Amazon S3 and a bunch of other different file systems and servers. 

To configure RIIIF in OD2 Follow these steps. 

* Add RIIIF gem to gemfile and bundle install and rebuild docker:
```
gem 'riiif', '~> 1.1' 
```

* Configure RIIIF in hyrax.rb initializer file
```
config.iiif_image_server = true
   # Returns a URL that resolves to an image provided by a IIIF image server
  config.iiif_image_url_builder = lambda do |file_id, base_url, size|
    "#{base_url}/downloads/#{file_id.split('/').first}"
  end
   # Returns a URL that resolves to an info.json file provided by a IIIF image server
  config.iiif_info_url_builder = lambda do |_, _|
    ""
  end
   # Returns a URL that indicates your IIIF image server compliance level
  config.iiif_image_compliance_level_uri = 'http://iiif.io/api/image/2/level2.json'
   # Returns a IIIF image size default
  config.iiif_image_size_default = '600,'
   # Fields to display in the IIIF metadata section; default is the required fields
  config.iiif_metadata_fields = Hyrax::Forms::WorkForm.required_fields
```

* Add RIIIF initializer file with proper configurations

```
Riiif::Image.file_resolver = Riiif::HTTPFileResolver.new
Riiif::Image.info_service = lambda do |id, _file|
  # id will look like a path to a pcdm:file
  # (e.g. rv042t299%2Ffiles%2F6d71677a-4f80-42f1-ae58-ed1063fd79c7)
  # but we just want the id for the FileSet it's attached to.

  # Capture everything before the first slash
  fs_id = id.sub(/\A([^\/]*)\/.*/, '\1')
  resp = ActiveFedora::SolrService.get("id:#{fs_id}")
  doc = resp['response']['docs'].first
  raise "Unable to find solr document with id:#{fs_id}" unless doc
  { height: doc['height_is'], width: doc['width_is'] }
end

Riiif::Image.file_resolver.id_to_uri = lambda do |id|
  ActiveFedora::Base.id_to_uri(CGI.unescape(id)).tap do |url|
    Rails.logger.info "Riiif resolved #{id} to #{url}"
  end
end

Riiif::Image.authorization_service = Hyrax::IIIFAuthorizationService

Riiif.not_found_image = Rails.root.join('app', 'assets', 'images', 'us_404.svg')
Riiif.unauthorized_image = Rails.root.join('app', 'assets', 'images', 'us_404.svg')

Riiif::Engine.config.cache_duration_in_days = 365
```

* Mount the engine in routes.rb file:
```
mount Riiif::Engine => 'images', as: :riiif if Hyrax.config.iiif_image_server?
```

revgum:
fixes #54 
ref #30 

This adds RAIS (without S3 integration) to the development/test stack, adds an OD FileSetDerivativesService to generate jp2's for tiffs and fullsize jpg's for other image types. Generates valid IIIF urls for RAIS to serve the images up using Universal Viewer. 

I couldn't find a great way to access the files directly on Fedora which is why full sized derivatives are being created.. This may make sense, in so much as the original files are stored in the backend (currently fedora) and RAIS has access to jp2 or full size jpg. Let's discuss.

See the screenshot with a gif, 2 png, 2 jpg, and a tiff original files.. The tiff was transcoded to a jp2 and the other files to full sized jpgs. 

screen shot 2018-11-23 at 4 29 55 pm


app/models/file_set.rb
```diff
@@ -1,4 +1,19 @@
 # Generated by hyrax:models:install
 class FileSet < ActiveFedora::Base
   include ::Hyrax::FileSetBehavior
+
+  ##
+  # Generate an encoded iiif url for this FileSet
+  # @param base_url [String] : provided by iiif_image_url_builder or iiif_info_url_builder, looks like http://localhost
+  # @return [String] : a url to the iiif server with a properly encoded identifier, looks like http://IIIFSERVER/ab%2fcd%2f12%2f24%2fz-jp2.jp2
+  def iiif_url(base_url)
```
jechols: For S3 storage, the IIIF identifier will just be "prefix" + "s3:" + encoded s3 identifier.  For instance, `https://rais/iiif/s3:mycollection%2Ffile.jp2`

Just an FYI, not a failure - I know what you have is good for a prototype.

jechols:
Oh, I suppose this would be a better place for discussion to capture our thoughts.  So here's a summary:

- Kakadu may not be sufficient since we probably want a JP2 for every image just for consistency.
  - It would be especially painful if we found something like a 50-megapixel PNG that RAIS ended up having to process.  RAIS *can* do this, but it's a lot more RAM and disk I/O than a tiled, multi-res JP2.
  - We could build a "X to TIFF" processor that happens prior to the JP2 processor, but it wouldn't be very efficient compared to just having a single processor that does it all.
- I'll create a ticket to build a Hyrax JP2 processor that uses GraphicsMagick, since that'll handle all file types.  I'll shoot for a smart processor based on my research in https://github.com/uoregon-libraries/rais-image-server/wiki/How-To-Encode-JP2s
- I'm of the opinion that we don't need full-size images if we have a high-quality JP2 for all images.  The value of a 100-megapixel JPEG, for instance, seems iffy.  And for other use-cases, a "medium" image is probably fine.
- Semi-related: I don't think we want to be storing any binary assets directly in Fedora.  The costs would be quite a bit higher than storing the bitstreams in S3.  S3-IA is something like 1/5th the cost of EBS, and 1/20th the cost of EFS.

app/models/file_set.rb
revgum: For RAIS, `s3:` tells it to use s3, but then do you include the full bucket path to the file? @jechols

spec/models/file_set_spec.rb
```diff
@@ -1,7 +1,18 @@
 # frozen_string_literal: true
 
 RSpec.describe FileSet do
-  it "has tests" do
-    skip "Add your tests here"
+  subject { create(:file_set) }
+  let(:encoded_id) { subject.id.scan(/.{1,2}/).join('%2f') }
+  let(:iif_url) { "#{ENV.fetch('IIIF_SERVER_BASE_URL', 'http://bogus')}/#{encoded_id}" }
```
jechols: Is it worth the trouble to fix this var?  (Should be `iiif_url`)

Issue #77, opened 2018-11-13
Error handling for oEmbed viewers


wickr:
### Descriptive summary

If a work has an oEmbed URL, and the viewer fails to load on the show page, that should be recorded and logged for troubleshooting. Possible reasons are that the source site is down, permissions changed on the media, or it was removed.

### Expected behavior

- [x] Admin Dashboard shows all oEmbed errors, sorted by most recent, and what those errors were.
- [x] Provide a way to mark errors as resolved.
- [ ] Send email to depositor of work with oEmbed URL notifying about error.

### Related work

#75 
#76

vantuyls:
@wickr is there QA required for this?

CGillen:
QA:
- [ ] Errors dashboard makes sense and is usable as expected from maintainer's perspective.
- [ ] Email is sent to depositor on error (or is attempted until emails work).

PR #148, opened 2018-11-27
Admin Dashboard view of oEmbed enabled items

### LINKED ISSUE #76: Admin Dashboard view of oEmbed enabled items
### Descriptive summary

Admins and staff should be able to easily see only all works that have an oEmbed URL included.

### Expected behavior

An Admin Dashboard view can show which works have an oEmbed URL included.

### Related work

#75 Initial configuration of oEmbed

CGillen:
Fixes #76


app/controllers/hyrax/my_controller.rb.bak
```diff
@@ -0,0 +1,67 @@
+module Hyrax
```
revgum: Don't forget to remove this file

CGillen:
Current dashboard layout as of this commit:
image

- Title links to the work: `http://localhost:3000/concern/generics/pr76f340k?locale=en`
- oEmbed Link links to the visible url
- oEmbed Status is dummy placeholder text to swap out for logic of #77

revgum:
Screenshot is looking good! Aside from the feedback you're asking from Ryan, this PR just needs tests to cover the code.

wickr:
@CGillen No, the work URI doesn't need to be in a separate column. The link in the Title should be sufficient.

There shouldn't be more than 1 oEmbed link for any given work, so you can drop the bulleted list to simplify things.

CGillen:
Then we should codify a single value relationship for the oEmbed URL. I can make that part of this PR for expedition.

PR #155, opened 2018-11-29
adds basic oauth stuff for login through cas

straleyb:


config/routes.rb
```diff
@@ -8,7 +8,12 @@
     concerns :searchable
   end
 
-  devise_for :users
+  devise_for :users, controllers: { omniauth_callbacks: "users/omniauth_callbacks" }
+  devise_scope :user do
+    get 'users/auth/cas', to: 'users/omniauth_authorize#passthru', defaults: { provider: :cas }, as: "new_osu_session"
+    get 'users/auth/shibboleth', to: 'users/omniauth_authorize#passthru', defaults: { provider: :cas }, as: "new_uo_session"
```
revgum: The provider is incorrect here.

app/views/devise/shared/_links.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <%= link_to "Didn't receive confirmation instructions?", new_confirmation_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.lockable? && resource_class.unlock_strategy_enabled?(:email) && controller_name != 'unlocks' %>
+  <%= link_to "Didn't receive unlock instructions?", new_unlock_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.omniauthable? %>
+  <%- resource_class.omniauth_providers.each do |provider| %>
+    <%= link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider) %><br />
```
revgum: Does this render links like "Sign in with Cas" and "Sign in with Saml"? If so, we would want to improve these to be some OSU and UO verbiage

config.ru
```diff
@@ -1,8 +1,8 @@
-# frozen_string_literal:true
+# This file is used by Rack-based servers to start the application.
+require_relative 'config/environment'
```
revgum: This needs to be loaded after Honeycomb, your PR shouldn't need to affect this file.

straleyb: This happened on the rebase. Ill update.

app/views/devise/shared/_links.html.erb
straleyb: Yes it does. Since ive overridden those pages, I can update this with some tricky code to translate a provider to a university.

app/models/user.rb
```diff
@@ -12,12 +12,28 @@ class User < ApplicationRecord
...
 
   # Method added by Blacklight; Blacklight uses #to_s on your
   # user class to get a user-displayable login/identifier for
   # the account.
   def to_s
     email
   end
+
+  def self.from_omniauth(access_token)
+    if access_token.provider.to_s == "cas"
```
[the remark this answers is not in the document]
straleyb: Ahh yeah, I see. So something like
```
user.where(:email => "blah") do |u|
     switch access_token.provider.to_s
     case "cas"
     case "saml"
     default
end
```

app/views/layouts/hyrax.html.erb
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
```
revgum: Is this being overwritten from the default provided by hyrax? If so, add the tag to comment where it differs?

config.ru
```diff
@@ -1,5 +1,3 @@
-# frozen_string_literal:true
```
revgum: Let's leave this in the file, we don't have rubocop to enforce style rules yet, but these should be included in all ruby code

app/views/_user_util_links.html.erb
```diff
@@ -0,0 +1,39 @@
+<ul id="user_utility_links" class="nav navbar-nav navbar-right">
```
revgum: If this is overwritten from Hyrax, can you comment where it differs?

app/views/_masthead.html.erb
```diff
@@ -0,0 +1,20 @@
+<header>
```
revgum: If this is overwritten from Hyrax, can you comment where it differs?

app/controllers/users/registrations_controller.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+require 'mail'
+
+class Users::RegistrationsController < Devise::RegistrationsController
+  before_action :redirect_if_university, only: [:create]
+
+  protected
+
+    def redirect_if_university
+      Rails.logger.info(Mail::Address.new(params[:user][:email]).domain.to_s)
```
revgum: set a `domain` variable rather than call `Mail::Address.new` twice (for the logging and then the case-statement).

Also, if this logging line is to be kept in the code, can you add some context about or detail of what is happening so the log includes something meaningful to read later.

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+require 'mail'
+
+class Users::SessionsController < Devise::SessionsController
+  before_action :redirect_if_university, only: [:create]
+
+  protected
+
+    def redirect_if_university
+      case Mail::Address.new(params[:user][:email]).domain.to_s
```
revgum: Log here like in the registrations_controller?

config/environments/development.rb
```diff
@@ -33,8 +33,21 @@
   # Don't care if the mailer can't send.
   config.action_mailer.raise_delivery_errors = false
 
+  config.action_mailer.default_url_options = { :host => 'localhost:3000' }
```
revgum: take host from an ENV, defaulting to `localhost:3000` .. `ENV.fetch('APPLICATION_URL', 'localhost:3000')`

config/environments/development.rb
```diff
@@ -33,8 +33,21 @@
...
+
   config.action_mailer.perform_caching = false
 
+  config.action_mailer.delivery_method = :smtp
+
+  config.action_mailer.smtp_settings = {
+  address: 'smtp.gmail.com',
```
revgum: I think we should/could expect SMTP to be active on the server or in the infrastructure rather than use gmail (I understand this was just for testing).. 

I'd like to see the development, production, and staging configurations default to using SMTP local to the application (assume that the server has smtp) and those configuration values tie to ENV but with localhost defaults.

config/initializers/devise.rb
```diff
@@ -253,7 +253,10 @@
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
   # config.omniauth :github, 'APP_ID', 'APP_SECRET', scope: 'user,public_repo'
-
+  config.omniauth :cas, host: 'login.oregonstate.edu', url: 'https://login.oregonstate.edu/cas'
+  config.omniauth :saml,
+    idp_cert: 'cert',
```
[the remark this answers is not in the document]
CGillen: Also, since we don't have a final URL for production yet, I can only submit for trusted access for staging. If URLs change, I'm pretty sure our trusted status breaks.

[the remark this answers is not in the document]
revgum: Good idea to move these strings to ENV variables @straleyb with defaults..

spec/controllers/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
revgum: require 'rails_helper' is no longer required or our specs

spec/models/user_spec.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
```
revgum: require 'rails_helper' is no longer required or our specs

.rubocop_fixme.yml
```diff
@@ -165,9 +165,18 @@ Metrics/BlockLength:
...
 
 Metrics/MethodLength:
   Exclude:
     - 'lib/oregon_digital/triple_powered_properties/inputs/triple_powered_property_input.rb'
     - 'lib/oregon_digital/triple_powered_properties/work_behavior.rb'
     - 'lib/tasks/oregon_digital/csv_ingest.rake'
+
+Style/SymbolArray:
+  Exclude:
+    - 'app/models/user.rb'
```
revgum: can you move this to the .rubocop.yml along with a comment of why the exclusion is needed

vantuyls:
@straleyb can  you please include descriptions in   your PRs?

PR #156, opened 2018-11-29
Added some additional metadata for requests

decimalator:


app/controllers/application_controller.rb
```diff
@@ -13,4 +13,10 @@ class ApplicationController < ActionController::Base
   skip_after_action :discard_flash_if_xhr
 
   protect_from_forgery with: :exception
+
+  def append_info_to_payload(payload)
```
revgum: I think this is a case of needing to surround this block in an if-statement to apply only to `staging` and `production`

Issue #113, opened 2018-11-16
Offload oEmbed URL fetching from the server


CGillen:
### Descriptive summary

The oEmbed currently renders the embed via the rails backend. This can cause the application to hang if the provider doesn't respond in a predicated way.

### Expected behavior

oEmbed content appears on the page with minimal chance of application stalling

### Related work

https://github.com/sul-dlss/blacklight-oembed may have done something similar
#75

CGillen:
Check PR #192 for current workflow. This is the method blacklight-oembed uses and seemed to make sense and was easy to implement.
If we still want to develop a rake task, that should be possible.

PR #179, opened 2018-12-05
adds rubocop

revgum:


app/renderers/oembed_attribute_renderer.rb
```diff
@@ -1,8 +1,12 @@
...
-      resource = OEmbed::Providers.get(value)
-      html << resource.html
-      return html
-    end
+  # TODO: Offload the loading of the OEmbed value to the client,
+  # render only the link to client-side.
+  def attribute_value_to_html(value)
+    resource = OEmbed::Providers.get(value)
+    resource.html
+  rescue OEmbed::NotFound
```
CGillen: Won't matter after #113 , but this breaks bubbling the error up out of the renderer into a context that can access the work controller/id to generate the OmbedError object.
Error is later rescued in `app/views/hyrax/base/_attribute_rows.html.erb` which is awful, but like I said, fixed in PR #192

CGillen: Non-issue. Fixed later.

PR #192, opened 2018-12-07
Offload oEmbed URL fetching from the server

### LINKED ISSUE #113: Offload oEmbed URL fetching from the server

CGillen:
Still fetched server side, but done after page loads.
Show pages loads => JS requests `/oembed/embed?url=<oembed_url>` => server requests oEmbed endpoint and returns JSON* `{ html : <html> }` => JS embeds response.

*If ruby-oembed encounters an error at this point an OembedError is generated for every work that oembed_url is found in and response is blank. JS doesn't embed anything.

fixes #113


app/controllers/blacklight/oembed/embed_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  # Blacklight later
+  class EmbedController < ActionController::Base
+    def show
+      render json: { html: get_embed_content(params[:url], additional_params) }
+    end
+
+    private
+
+    def get_embed_content(url, add_params)
+      # rubocop:disable Style/RedundantBegin
```
revgum: Remove the `begin` and the rubocop override. 

```
def method
  do_something
rescue 
  do_something_else
end
```

app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -21,9 +21,4 @@
 <% OregonDigital::GenericMetadata::PROPERTIES.map(&:to_sym).each do |attr| %>
```
revgum: Add a `.reject` to remove `oembed_url` from the resulting array since the rendering happens below.

app/controllers/blacklight/oembed/embed_controller.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Blacklight::Oembed
+  # Override blacklight-oembed gem controller. We're only interested in
+  # get_embed_content, but keep the changes compatible incase we can use
+  # Blacklight later
+  class EmbedController < ActionController::Base
+    def show
+      render json: { html: get_embed_content(params[:url], additional_params) }
```
straleyb: I think instead of calling these `additional_params`, the rails way of doing is is calling them `options` or `opts`

app/controllers/blacklight/oembed/embed_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+        OembedError.find_or_create_by(document_id: work.id).add_error(e)
+      end
+
+      "<dt>oEmbed encounted an error</dt><dd>#{e.message}</dd>".html_safe
+    end
+
+    ##
+    # Allows for blacklight-oembed users to pass through additional configured
+    # parameters
+    def additional_params
```
straleyb: Could call this method oembed_options

app/controllers/blacklight/oembed/embed_controller.rb
CGillen: Overridden from: https://github.com/sul-dlss/blacklight-oembed/blob/master/app/controllers/blacklight/oembed/embed_controller.rb

I tried to keep it as backwards compatible as possible in case Hyrax decides to make use of Blacklight show pages in the future. If so, we can reconfigure this to work fairly easily.

straleyb: Cool, can we make a note of that with a comment then? Just for later down the road when no one has looked at the code, we will know this was overridden at some point from somewhere.

app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -1,4 +1,4 @@
-<% OregonDigital::GenericMetadata::PROPERTIES.map(&:to_sym).each do |attr| %>
+<% OregonDigital::GenericMetadata::PROPERTIES.map(&:to_sym).reject{|attr| attr==:oembed_url}.each do |attr| %>
```
revgum: I hope this doesn't sound nitpicky.. but since rubocop doesn't check views, it wouldn't point this out.. can you make the `reject` a little more readable

`.reject { |attr| attr == :oembed_url }`

CGillen: Not at all. Also, I forgot to check rubocop anyways.
Need to get more in the habit

PR #198, opened 2018-12-07
Errored oembed list update

### LINKED ISSUE #197: Manage oEmbeds dashboard - oEmbed errors tab
### Descriptive summary

oEmbed errors tab is rendering an unordered list of errors and the style isn't matching the "All oEmbed Works" tab. Here are some updates that I think would finish this view off.

### Expected behavior

- [ ] Update tab title to 'oEmbed Errors', note the capital E in Errors
- [ ] Use `_list_oembeds.html.erb` partial as the basis for `_list_errored_oembeds.html.erb` but render the list with `assets_with_errored_oembed` instead.

revgum:
fixes #197 

screen shot 2018-12-07 at 3 28 54 pm
screen shot 2018-12-07 at 3 28 47 pm


CGillen:
Are we moving the interface this way? Previously, the design was parallel to embargo and lease dashboards:
image
If we go this way, it seems redundant to have different tabs for all works and errored works.

revgum:
@CGillen Good question, I was just trying to make the tabs look consistent (the deactiveated embargoes tab looks bad IMO).. I think the oEmbed Errors is handy in that the user will get the filtered with the errors versus a big list of all oEmbeds in the system.

CGillen:
Agreed that the default interface looks bad. I guess it's safe to assume that and maybe this one will be changed multiple times anyways.
It seems like adding sortable columns or an in-table filter would accomplish the same thing, but adding a tab is probably the easy solution.
Rebase and I'll approve+merge.

Issue #87, opened 2018-11-13
Generate and config Generic work type


zhang4952:
### Descriptive summary

We need to generate files for `Generic` work type for OD2 using a rake task command provided by Hyrax, and customize the generated files based on OD2 MAP (https://docs.google.com/spreadsheets/d/16xBFjmeSsaN0xQrbOpQ_jIOeFZk3ZM9kmB8CU3IhP2c/edit#gid=0)
- actor
- controller
- indexer
- forms
- presenters
- models
- views
- config
- spec

### Expected behavior

Create the files of `Generic` work type should be straightforward. The customization/configuration of created files could require quite some efforts. 

### Related work

OD2 MAP: https://docs.google.com/spreadsheets/d/16xBFjmeSsaN0xQrbOpQ_jIOeFZk3ZM9kmB8CU3IhP2c/edit#gid=0

Link to related tickets or prior related work here.
#21 
#27 

### Steps

- [x] Generate files with Hyrax rake task

- [x] Check the completeness of MAP with `Generic`

- [x] Configure `Generic` files with metadata fields

luisgreg99:
@zhang4952 It looks like the first step (rake task to generate default files such as model, controller, actors) for both Generic and Image models has been covered by https://github.com/OregonDigital/OD2/pull/67 I think we can built up on those files to extend or customize the metadata.

zhang4952:
#103

revgum:
Has this been completed @zhang4952 @straleyb

zhang4952:
I think it has been merged and in `QA` status now.

wickr:
Addressed by #129 and partly #153

wickr:
QA Issues from #129 

- [ ] :license in Generic metadata twice - Remove  property :license, predicate: ::RDF::Vocab::CC.License
- [ ] :source in Generic metadata twice - Remove 1st occurrence
- [ ] :has_part in Generic metadata twice - Remove 2nd occurrence (with :facetable)
- [ ] :has_part in PROPERTIES twice
- [ ] SpecimenType can use '_' in property name
- [ ] :identification_verification_status in PROPERTIES twice
- [ ] capitalized 'O' in predicate: property :identification_verification_status, predicate: ::RDF::Vocab::DWC.identificationVerificatiOnStatus
- [ ] :creator not in PROPERTIES
- [ ] :contributor not in PROPERTIES
- [ ] :description not in PROPERTIES
- [ ] :rights_statement not in PROPERTIES
[...]

PR #216, opened 2018-12-12
Updates metadata per QA

straleyb:
This PR addresses the fixes requested in #87


straleyb:
Date created, related url, and bibliographic citation are all required by the app.

app/forms/hyrax/generic_form.rb
```diff
@@ -11,7 +11,7 @@ class GenericForm < Hyrax::Forms::WorkForm
     self.required_fields = %i[title dcmi_type rights_statement]
 
     def primary_terms
-      required_fields + (OregonDigital::GenericMetadata::PROPERTIES.map(&:to_sym) - [:dcmi_type])
+      required_fields + (OregonDigital::GenericMetadata::PROPERTIES.map(&:to_sym) - required_fields)
```
wickr: You could just remove the required_fields here right? instead of adding and then removing them

app/indexers/generic_indexer.rb
```diff
@@ -3,11 +3,11 @@
 class GenericIndexer < Hyrax::WorkIndexer
   # This indexes the default metadata. You can remove it if you want to
   # provide your own metadata and indexing.
-  include Hyrax::IndexesBasicMetadata
+  # include Hyrax::IndexesBasicMetadata
```
wickr: Go ahead and delete these.

app/forms/hyrax/generic_form.rb
straleyb: No this makes sure that the required fields end up at the top. Otherwise they would be in the middle somewhere. For ease of testing.

revgum: Can you be sure to make a comment about that so it's understandable "why"?

wickr: Oh, I get it now, I wasn't seeing all the parentheses earlier. Thanks.

PR #221, opened 2018-12-17
oEmbed error email notifications

### LINKED ISSUE #77: Error handling for oEmbed viewers

CGillen:
Creates new Hyrax notifications for depositor when a new error is added.

@wickr The original ticket states it should be emails. Are we expecting depositors to rarely come back to OD unless they get an external ping? If so, I should expand this to use ActionMailer.

fixes #77


wickr:
@CGillen I think Hyrax notifications are good but not sufficient on their own, and an email is the overall best way to let the depositor know something is wrong. Some depositors may return regularly, but others won't. 

Hyrax has some infrastructure for emails, particularly around workflow updates, and we customized this some with ScholarsArchive@OSU. Maybe @gum @straleyb (also @luisgreg99 but he's away for a couple weeks) have some more thoughts on how to integrate emails with what's already there.

CGillen:
It looks like ScholarsArchive reconfigured mailboxer to send emails w/ Hyrax notifications. Should we do the same for OD2? That would solve the issue here and send more emails for other notifications. I'm betting workflows sends notifications too, so that would solve sending emails for workflow changes.

@vantuyls Is this a PMOP/stakeholders discussion?

CGillen:
Opened #230 to determine if notifications is sufficient here or if we need to build out more robust email infrastructure or make this a one-off email notification (not Hyraxy)

app/models/oembed_error.rb
```diff
@@ -27,6 +28,13 @@ def add_error(error)
...
   def unique_errors
     oembed_errors.map!(&:to_s).uniq!
-    touch if persisted?
+  end
+
+  def alert_depositor
+    # Only alert user if new errors are added
+    return unless saved_change_to_oembed_errors?
+
+    user = User.find_by_user_key(Generic.find(document_id).depositor)
```
revgum: It might be better, here, to query Fedora directly:
`user = User.find_by_user_key(Generic.find(document_id).depositor)`
to 
`user = User.find_by_user_key(ActiveFedora::Base.find(document_id).depositor)`

Issue #190, opened 2018-12-07
Make list of supported oEmbed external sources


wickr:
### Descriptive summary

We should make a list of supported oEmbed sources that we can confirm will work in the system.

Include: OSU MediaSpace (Kaltura), YouTube, Vimeo, others?

### Related work

#75

CGillen:
The ruby-oembed gem wins us these providers for free: https://github.com/ruby-oembed/ruby-oembed/blob/master/lib/oembed/providers.rb#L139-L359
Plus we can add oohEmbed, noEmbed, and Embedly aggregators.
After that, we need to follow the instructions for adding custom providers: https://github.com/ruby-oembed/ruby-oembed#custom-providers

wickr:
Keep providers included by plugin, especially YouTube and Vimeo
Add OSU MediaSpace (Kaltura)

vantuyls:
QA PASS - whitelisted (youtube) oembed works
QA PASS -  non-whitelisted (daily motion) oembed fails

PR #232, opened 2018-12-18
Add OSU MediaSpace provider

### LINKED ISSUE #191: oEmbed URLs are checked against a whitelist of sources/domains
### Descriptive summary

We don't particularly want to support any available oEmbed source, but a few major ones. Plus a whitelist helps us confirm that they work as expected.

### Expected behavior

oEmbed URLs not recognized should be refused, with note to contact project team to look into including as an additional source.

Recognized oEmbed URLs should perform as expected.

### Related work
#75 
#190

CGillen:
Fixes #191 

As discussed in #191, whitelisting is a part of ruby-oembed. Adding the OSU MediaSpace as a provider adds the last identified (#190) provider to the whitelist.


revgum:
@CGillen is it possible to write a test to verify/validate that the mediaspace provider has been added as a provider?

PR #235, opened 2018-12-19
[WIP] Fixes pieces of indexing

straleyb:
This refines the indexing a bit more. Since we have our own custom metadata with some core pieces of metadata taken out, we need to remove a bit more of the hyrax work in order to replace it with our. 

IndexesLinkedMetadata
IndexesBasicMetadata
MetadataIndexer
DeepIndexingService

Are ones to look at. We had to take these out because they were trying to index values that we dont support anymore.

### DIFF
--- a/app/indexers/concerns/oregon_digital/indexes_basic_metadata.rb
+++ b/app/indexers/concerns/oregon_digital/indexes_basic_metadata.rb
@@ -1,0 +1,9 @@
+module OregonDigital
+  # This module can be mixed in on an indexer in order to index the basic metadata fields
+  module IndexesBasicMetadata
+    # We're overriding a method from ActiveFedora::IndexingService
+    def rdf_service
+      OregonDigital::MetadataIndexer
+    end
+  end
+end
--- a/app/indexers/generic_indexer.rb
+++ b/app/indexers/generic_indexer.rb
@@ -1,18 +1,18 @@
 # frozen_string_literal:true
 
 class GenericIndexer < Hyrax::WorkIndexer
-  # This indexes the default metadata. You can remove it if you want to
-  # provide your own metadata and indexing.
-
-  # Fetch remote labels for based_near. You can remove this if you don't want
-  # this behavior
+  include OregonDigital::IndexesBasicMetadata
+  include OregonDigital::IndexesLinkedMetadata
 
   # Uncomment this block if you want to add custom indexing behavior:
   def generate_solr_document
     super.tap do |solr_doc|
       OregonDigital::GenericMetadata::PROPERTIES.map(&:to_s).each do |prop|
-        solr_doc["#{prop}_ssim"] = object.attributes[prop.to_s]
-        solr_doc["#{prop}_tesim"] = object.attributes[prop.to_s]
+        if !object.attributes[prop.to_s].is_a? ActiveTriples::Relation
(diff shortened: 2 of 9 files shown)


app/indexers/generic_indexer.rb
```diff
@@ -1,19 +1,30 @@
...
-  # Fetch remote labels for based_near. You can remove this if you don't want
-  # this behavior
-
-  # Uncomment this block if you want to add custom indexing behavior:
   def generate_solr_document
     super.tap do |solr_doc|
-      OregonDigital::GenericMetadata::PROPERTIES.map(&:to_s).each do |prop|
-        solr_doc["#{prop}_ssim"] = object.attributes[prop.to_s]
-        solr_doc["#{prop}_tesim"] = object.attributes[prop.to_s]
+      OregonDigital::DocumentMetadata::PROPERTIES.map(&:to_s).each do |prop|
```
CGillen: `OregonDigital::GenericMetadata::PROPERTIES.map(&:to_s).each do |prop|`

app/indexers/generic_indexer.rb
```diff
@@ -1,19 +1,30 @@
...
+        else
+          index_value_for_singular(solr_doc, attr, prop)
+        end
       end
     end
   end
+
+  private
+
+  def index_value_for_multiple(solr_doc, attr, prop)
```
CGillen: Can these be factored out of the other metadata indexers?

app/indexers/oregon_digital/metadata_indexer.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module OregonDigital
+  # OVERRIDE FROM HYRAX TO ADD OUR OWN FIELDS
+  class MetadataIndexer < ActiveFedora::RDF::IndexingService
```
CGillen: Could this be simplified by inheriting from `Hyrax::BasicMetadataIndexer` then just overriding the bits that we actually want to change? Is this better or worse practice in Ruby/Rails land?

app/indexers/audio_indexer.rb
```diff
@@ -1,18 +1,8 @@
...
-
-  # Uncomment this block if you want to add custom indexing behavior:
-  # def generate_solr_document
-  #  super.tap do |solr_doc|
-  #    solr_doc['my_custom_field_ssim'] = object.my_custom_property
-  #  end
-  # end
+class AudioIndexer < GenericIndexer
+  def generate_solr_document
+    super.tap do |solr_doc|
```
CGillen: Doesn't metadata still exist from generic that needs to be generated in solr here?

straleyb: Yeah, when it calls super, it uses the GenericIndexer code and generates all that.

app/indexers/concerns/oregon_digital/indexes_basic_metadata.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module OregonDigital
```
CGillen: The changes in this PR removed references to IndexesBasicMetadata and IndexesLinkedMetadata, so do we still need these overrides?
If so, since these are under the OregonDigital module, wouldn't the references in Hyrax not see these overrides unless they're in the Hyrax module?

PR #246, opened 2018-12-20
Feature/remove dcmi type

### LINKED ISSUE #234: Remove dcmi_type property from GenericMetadata
### Descriptive summary

We will be sticking with Resource Type set to DC.type, and updating the authority values as desired. We don't need a separate property of `dcmi_type`, and removing it will eliminate the predicate conflict. Resource Type can stay required.

### Related work

#233 Resource Type changing to DCMI values

straleyb:
Fixes #234


app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -412,9 +412,9 @@ module GenericMetadata
         index.as :stored_searchable, :facetable
       end
 
-      property :dcmi_type, predicate: ::RDF::Vocab::DC.type, multiple: false do |index|
-        index.as :stored_searchable, :facetable
-      end
+      # property :dcmi_type, predicate: ::RDF::Vocab::DC.type, multiple: false do |index|
```
revgum: Remove the property instead of leaving commented out

PR #250, opened 2018-12-21
configure staging and production for smtp

### LINKED ISSUE #248: [1] Configure proper URL to be sent out when user tries to recover password
### Descriptive summary

Currently, when a user requests to recover their password, the email that gets sent out includes a url that links to `localhost`. This is because it is not configured to use the actual url the application is at. This means on staging it should route to `od2-staging.library.oregonstate.edu`. Once this is configured, this should allow users to request their password on staging. 

### Current functionality

When a user requests their password, the link that is sent out goes to `localhost`

### Expected functionality

When a user requests their password, the link that is sent out goes to `od2-staging.library.oregonstate.edu`

straleyb:
Fixes #248

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -34,7 +34,7 @@ Rails.application.configure do
 
   config.action_mailer.raise_delivery_errors = false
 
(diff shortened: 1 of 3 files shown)


revgum:
Not necessary for production/staging, these are managed outside of this repo

Issue #154, opened 2018-11-29
#show metadata fields 


revgum:
### Descriptive summary

Metadata field titles need to be set properly on the show page. An example of the current labels can be seen on http://od2-staging.library.oregonstate.edu/concern/documents/3r074t92z?locale=en#?c=0&m=0&s=0&cv=0&xywh=-490%2C0%2C2179%2C750

See the `form label` field in the MAP, could be that the field title just needs a `.humanize`

mccallum:
I can't comment. http://od2-staging.library.oregonstate.edu/ isn't working for my earlier @uoregon.edu login... I've tried Incognito mode and clearing cache but it just flips to Shib saying "no." I created a non-institutional @maymarch.com account and I get logged in but showpage presents a "We're sorry, but something went wrong. If you are the application owner check the logs for more information."

mccallum:
We were asked to leave comments on things in QA column to possibly clear and this was labeled UX.

vantuyls:
@jsimic can you or another metadataoid QA this something this week?

jsimic:
@vantuyls Sure

simholt:
Reopening this -- see spreadsheet below for details. 

https://docs.google.com/spreadsheets/d/1y27Gtcm42qG7mKQTVNhkOIoPTD9viW65n7vhkI2IPGE/edit?usp=sharing

@jsimic Please look over spreadsheet. Some clarification needed about Type vs. Resource Type.

mccallum:
I see why this is labeled UX now.

Metadeities (@simholt @jsimic ) can add a "Top 15-30" column to their spreadsheet above for the metadata. Then we'll expose a full 160 plus possible with a "See all +" make-visible button item or something.  

Or divide them into four user-friendly categories as NU did: https://www.fulcrum.org/concern/file_sets/zc77sq117?locale=en#info

Or both.

straleyb:
Im pretty sure this ticket was for the configuration of the show page, once we get the order in place (if that is even necessary for this ticket), I think maybe this one should be good to go. Then we can make a new ticket for look and feel for the show page, rather than displaying proper info,

mccallum:
That makes sense, Brandon. I'm going to remove UX label from this and will remember my comment to @simholt and @jsimic for some future discussion since it got no response. Thanks.

wickr:
Confirming what's on Sarah's spreadsheet, these are current issues:

[...]

Issue #290, opened 2019-01-15
Image failed to save


revgum:
### Descriptive summary

On `Staging`, an Image is unable to save and fails with the following exception

```
I, [2019-01-15T22:52:37.119474 #33]  INFO -- : [7e41de89-681b-409e-977d-554cfe95fd02] Completed 500 Internal Server Error in 231ms (ActiveRecord: 12.8ms)
F, [2019-01-15T22:52:37.122411 #33] FATAL -- : [7e41de89-681b-409e-977d-554cfe95fd02]
F, [2019-01-15T22:52:37.122599 #33] FATAL -- : [7e41de89-681b-409e-977d-554cfe95fd02] ActiveFedora::UnknownAttributeError (unknown attribute 'date_created' for Image.):
F, [2019-01-15T22:52:37.127253 #33] FATAL -- : [7e41de89-681b-409e-977d-554cfe95fd02]
F, [2019-01-15T22:52:37.127404 #33] FATAL -- : [7e41de89-681b-409e-977d-554cfe95fd02] active-fedora (11.5.4) lib/active_fedora/attribute_assignment.rb:50:in `_assign_attribute'
[7e41de89-681b-409e-977d-554cfe95fd02] active-fedora (11.5.4) lib/active_fedora/attribute_assignment.rb:45:in `block in _assign_attributes'
[7e41de89-681b-409e-977d-554cfe95fd02] active-fedora (11.5.4) lib/active_fedora/attribute_assignment.rb:44:in `each'
[7e41de89-681b-409e-977d-554cfe95fd02] active-fedora (11.5.4) lib/active_fedora/attribute_assignment.rb:44:in `_assign_attributes'
[7e41de89-681b-409e-977d-554cfe95fd02] active-fedora (11.5.4) lib/active_fedora/attribute_assignment.rb:38:in `assign_attributes'
[7e41de89-681b-409e-977d-554cfe95fd02] active-fedora (11.5.4) lib/active_fedora/attribute_assignment.rb:9:in `attributes='
[7e41de89-681b-409e-977d-554cfe95fd02] active-fedora (11.5.4) lib/active_fedora/attributes/serializers.rb:24:in `attributes='
[7e41de89-681b-409e-977d-554cfe95fd02] /usr/local/bundle/bundler/gems/hyrax-c0699505fc89/app/actors/hyrax/actors/base_actor.rb:72:in `apply_save_data_to_curation_concern'
[...]

Issue #211, opened 2018-12-11
Remove Hyrax Core Metadata fields from GenericMetadata model


wickr:
### Descriptive summary

The Hyrax Core Metadata fields were added to GenericMetadata but they don't need to be there, they're still included. There are errors for the same predicates being defined for multiple instances of those properties.

### Expected behavior

We can rely on the Hyrax Core metadata fields being in place and don't need to specify them in GenericMetadata.

### Related work

#129

straleyb:
QA TESTING: This can be checked by cross referencing the code changes to the MAP. This was mostly a change to remove a lot of duplicate predicates which is fixed in #219

wickr:
Moving to In Progress since this is fixed by #219 which isn't yet merged.

wickr:
QA OK.

PR #292, opened 2019-01-16
updates per QA

straleyb:
Adresses QA in #154


app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -15,6 +15,13 @@ module GenericMetadata
       property :resource_type, predicate: ::RDF::Vocab::DC.type
       property :creator, predicate: ::RDF::Vocab::DC11.creator
       property :contributor, predicate: ::RDF::Vocab::DC11.contributor
+      property :date_created, predicate: ::RDF::Vocab::DC.created
+      property :date_modified, predicate: ::RDF::Vocab::DC.modified, multiple: false
```
wickr: date_modified and date_uploaded should be included via Hyrax core metadata and not defined again here.

app/indexers/generic_indexer.rb
```diff
@@ -9,7 +9,9 @@ def generate_solr_document
     super.tap do |solr_doc|
       OregonDigital::GenericMetadata::PROPERTIES.map(&:to_s).each do |prop|
         attr = object.attributes[prop]
-        if attr.is_a? ActiveTriples::Relation
+        if %w[date_created date_uploaded].include? prop
+          solr_doc["#{prop}_tesim"] = [Date.today] if object.attributes[prop].nil?
```
wickr: I don't think we should be setting values for these Hyrax-included properties, it should already be handled with things like https://github.com/samvera/hyrax/blob/master/lib/hyrax/transactions/steps/set_modified_date.rb

app/models/concerns/oregon_digital/generic_metadata.rb
straleyb: Since we heavily modified everything, we had to strip out all the core metadata, so we added it back in here. See resource_type, creator, contributor, etc. thats where the error from the ticket #290 comes from

app/indexers/generic_indexer.rb
straleyb: Same here. We heavily customized everything, so we needed to rebuild the indexer. This is what drives the whole indexing now and it ignores the hyrax base stuff because it expects core metadata we removed.

wickr: We stripped out Core metadata? Where? We just had a ticket to not do that: https://github.com/OregonDigital/OD2/issues/211

Even if we do, I don't know that the indexer should actually be setting the values. Is `Date.today` the exact same format that Hyrax is setting?

app/models/concerns/oregon_digital/generic_metadata.rb
wickr: Are you talking about we removed basic_metadata? resource_type, creator, etc. aren't in core_metadata

straleyb: Yeah, my bad. I was totally thinking about basic_metadata. Sometimes those seem like core to me. Momentary lapse of reasoning.

.rubocop_fixme.yml
```diff
@@ -14,6 +14,7 @@ Metrics/LineLength:
 Metrics/AbcSize:
   Exclude:
     - 'app/indexers/oregon_digital/deep_indexing_service.rb'
+    - 'app/indexers/generic_indexer.rb'
```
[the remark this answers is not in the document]
straleyb: Ill give it a go.

app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -1,4 +1,4 @@
-<% OregonDigital::GenericMetadata::PROPERTIES.map(&:to_sym).reject{ |attr| attr == :oembed_url }.each do |attr| %>
+<% (OregonDigital::GenericMetadata::PROPERTIES.map(&:to_sym) - [:description, :license]).reject{ |attr| attr == :oembed_url }.each do |attr| %>
```
revgum: don't use `-`, add those to the `reject` evaluation

straleyb:
Gotta hold off on this PR, there is an issue with how things are saving with these indexer changes. Forgot about the issue of values not being saved correctly I.E description not being saved as an empty array.

wickr:
I still don't think we should be indexing empty values in Solr but will make another ticket for that.

PR #301, opened 2019-01-18
monkeypatching hyrax, moving hacks

### LINKED ISSUE #298: BrowseEverything fails unexpectedly for S3
### Descriptive summary

S3 bucket has an image file in it, and the UI shows and allows to pick the image, but when it is tried to be imported by a background job it fails. Testing the S3 URL directly successfully fetched the image from S3.. so I'm not sure why "Error: Expired URL" is happening. Are these urls only single use?

See ticket #297 for the error that was attempting to be emailed (but email failed to send too).

revgum:
fixes #298 

Moves all local hacks to a more obvious location inside initializers


config/initializers/hacks/hyrax_create_with_remote_files_ordered_members_actor.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal:true
+
+Hyrax::Actors::CreateWithRemoteFilesActor.class_eval do
```
CGillen: `Hyrax::Actors::CreateWithRemoteFilesOrderedMembersActor`

PR #311, opened 2019-01-25
Feature/media type controlled

straleyb:


config/initializers/hyrax.rb
```diff
@@ -112,10 +112,8 @@
 
   # Location autocomplete uses geonames to search for named regions
   # Username for connecting to geonames
-  # config.geonames_username = ''
+  config.geonames_username = 'etsdev'
 
-  # Should the acceptance of the licence agreement be active (checkbox), or
```
wickr: Were these 2 lines meant to be deleted? (117 and 118)

app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -6,7 +6,8 @@ module GenericMetadata
...
+    CONTROLLED = %w[format_label]
```
wickr: There's got to be a better way to specify this. `format_label` isn't a property like the others. The property itself has the `class_name` attribute, and there's also `self.controlled_properties` where format is included. Can one of those be used for the access later?

app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -574,12 +571,14 @@ module GenericMetadata
       end
 
       property :based_near, predicate: ::RDF::Vocab::FOAF.based_near, class_name: Hyrax::ControlledVocabularies::Location
+      property :format, predicate: ::RDF::Vocab::DC.format, class_name: OregonDigital::ControlledVocabularies::MediaType
```
wickr: Why was the `index.as` line removed from :format ?

lib/oregon_digital/controlled_vocabularies/media_type.rb
```diff
@@ -0,0 +1,14 @@
+module OregonDigital
+  module ControlledVocabularies
+    class MediaType < ActiveTriples::Resource
+      configure rdf_label: ::RDF::Vocab::DC.format
```
wickr: This isn't right. I'm not positive you even need to define `rdf_label` at all here, it's an ActiveTriples::Resource already. Plus you're defining the label as the predicate URI, which is circular.

lib/oregon_digital/feature_class_uri_to_label.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module OregonDigital
+  # Translate feature class
+  class FeatureClassUriToLabel
+    def uri_to_label(uri)
+      { 'http://www.geonames.org/ontology#A' => 'Administrative Boundary',
+        'http://www.geonames.org/ontology#H' => 'Hydrographic',
+        'http://www.geonames.org/ontology#L' => 'Area',
+        'http://www.geonames.org/ontology#P' => 'Populated Place',
+       'http://www.geonames.org/ontology#R' => 'Road / Railroad',
```
wickr: Indent off starting here.

lib/oregon_digital/controlled_vocabularies/media_type.rb
straleyb: I did this because thats how location is done. https://github.com/samvera/hyrax/blob/5a9d1be16ee1a9150646384471992b03aab527a5/lib/hyrax/controlled_vocabularies/location.rb#L4 The RDF label I think is still able to be expressed as an ntriple which wants a predicate. But if thats not the case, then I can explore not configuring a specific predicate for an rdf_label for a specific controlled vocabulary

app/models/concerns/oregon_digital/generic_metadata.rb
straleyb: I removed it because I was following the way hyrax had done it previously. https://github.com/samvera/hyrax/blob/5a9d1be16ee1a9150646384471992b03aab527a5/app/models/concerns/hyrax/basic_metadata.rb#L29 I can try and add it back and see how it effects the system. I imagine it wont, so it should be cool.

app/models/concerns/oregon_digital/generic_metadata.rb
straleyb: This definitely wasn't the "Final solution" for this piece of code. This was to make it flow naturally through the rest of our code with minimal changes. But we should have a discussion about controlled vocabularies on the show page. Do we want them to be displayed as the search and external links? Do we want this to be different for each? We can add format_label to the properties which would be totally fine, but we need to decide how we would like to display these fields on the show page.

.rubocop_fixme.yml
```diff
@@ -20,6 +21,7 @@ Metrics/AbcSize:
     - 'lib/oregon_digital/triple_powered_properties/work_behavior.rb'
     - 'lib/tasks/oregon_digital/csv_ingest.rake'
     - 'spec/support/helpers/session_helpers.rb'
+    - 'lib/oregon_digital/controlled_vocabularies/extended_location.rb'
```
revgum: I think this should be addressed in the code instead of overridden and hidden

straleyb: I just put it here so I could work on getting tests ironed out. I didnt want a need to refactor to block me from getting tests passing. Ill make a note next time.

revgum: I'm curious if you're able to run these tests locally instead of testing by way of CircleCI? It should be a lot faster to run them locally..but I might not understand if there's a blocker preventing you from doing that?

straleyb: More like a not fully fleshed out development habit. From what I understand, to run the tests you have to bring docker down, spin it up in testing mode, run the tests, then to fire the app up, you have to switch back over right?

app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -6,7 +6,8 @@ module GenericMetadata
...
..._latitude gps_longitude has_finding_aid has_part has_version higher_classification hydrologic_unit_code identification_verification_status identifier illustrator inscription institution interviewee interviewer isPartOf is_version_of issued item_locator keyword language layout legal_name license local_collection_id local_collection_name location location_copyshelf_location longitude_latitude_identification lyricist material measurements military_branch military_highest_rank military_occupation military_service_location mode_of_issuance mods_note object_orientation oembed_url order original_name_usage owner patron photographer phylum_or_division physical_extent place_of_production primary_set print_maker provenance publication_place publisher ranger_district recipient relation replaces_url repository resource_type rights_holder rights_statement scientific_name_authorship scribe series_name series_number set source source_condition species specimen_type sports_team state_or_edition street_address style_or_period subject taxon_class technique temporal tgn transcriber translator tribal_classes tribal_notes tribal_terms tribal_title use_restrictions view_date water_basin workType].freeze
```
CGillen: date_created disappeared from this list.
This needs to be broken into multiple lines so we can diff it easier.

app/authorities/qa/authorities/format.rb
```diff
@@ -0,0 +1,30 @@
...
+    self.label = lambda do |item|
+      [item.first['http://www.w3.org/2000/01/rdf-schema#label'].first['@value']].compact.join(', ')
+    end
+
+    def search(q)
+      parse_authority_response(json(build_query_url(q)))
+    end
+
+    def build_query_url(q)
+      CGI.escape(q)
```
CGillen: CGI breaks for me, URI is fine

app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -6,7 +6,8 @@ module GenericMetadata
...
+    CONTROLLED = %w[format_label].freeze
```
CGillen: based_near_label?

app/authorities/qa/authorities/format.rb
CGillen: Better yet, allowing users to arbitrarily force us to hit any uri doesn't seem smart.
This should become:
```
query = CGI.escape(q)
"https://w3id.org/spar/mediatype/#{q}"
```
Then users only input `image/tiff` at the form and their autocomplete comes up, and we're guaranteed to hit the right vocab.

CGillen: CGI is still overbearing with escaping slashes. Also, QA doesn't automatically retrieve redirects. I had to solve the latter for some work with another vocab, so I'll push up my fix for that and will include something for slashes.

PR #331, opened 2019-02-06
Solr improvment

revgum:
eliminates the solr-cloud and configuration zip file upload process.. this creates the core if it doesn't exist.


revgum:
`docker-compose down -v --rmi local` will bring down the services, kill the volumes, and remove the local images so you can start all over

PR #337, opened 2019-02-07
Simple Triplestore Persistence 

straleyb:
For each ControlledVocabulary object that gets created, when persist! gets called on it, it takes the rdf_source which is a uri and commits it to the triplestore. Need to include OregonDigital::ControlledVocabularies::TriplestorePersistable which adds in the functionality of persisting to the triplestore.


lib/oregon_digital/controlled_vocabularies/resource.rb
```diff
@@ -32,6 +31,19 @@ def solrize
         [rdf_subject.to_s, { label: "#{rdf_label.first}$#{rdf_subject}" }]
       end
 
+      def persist!
+        triplestore_fetch
+        @persisted = true
+      end
+
+      def triplestore_fetch
+        URI.parse(rdf_subject).is_a?(URI::HTTP) ? triplestore.fetch(rdf_subject) : RDF::Graph.new
```
revgum: Should this be setting the subject on the `RDF::Graph.new`?

lib/oregon_digital/controlled_vocabularies/resource.rb
```diff
@@ -40,9 +63,9 @@ def rdf_label_uri_same?
 
       def uri_in_vocab?(uri)
         self.class.respond_to?(:in_vocab?) && self.class.in_vocab?(uri)
+        true
```
CGillen: Remove

app/authorities/qa/authorities/format.rb
```diff
@@ -24,8 +24,22 @@ def search(q)
 
     # Reformats the data received from the service
     def parse_authority_response(response)
-      [{ 'id' => response.first['@id'].to_s,
-         'label' => label.call(response) }]
+      id = find_response_id(response)
```
CGillen: I would prefer if QA Authorities used as similar as possible methods for finding the correct term.

This should work for every vocab that will only use URIs:
https://github.com/OregonDigital/OD2/blob/3b12af3604426842fa0d2c7ac93d885d050e34e3/app/authorities/qa/authorities/ethnographic_term.rb#L31-L33
Maybe it should be pushed up to Resource and overriden later.

We'll want a different strategy for label searches and maybe multiple vocabularies.

lib/oregon_digital/controlled_vocabularies/resource.rb
```diff
@@ -32,6 +40,21 @@ def solrize
         [rdf_subject.to_s, { label: "#{rdf_label.first}$#{rdf_subject}" }]
       end
 
+      def persist!
```
CGillen: I'm concerned about the parent/repository persistence strategy funny business that happens in ActiveTriples. I think this persists to Fedora at some point, are we sure that doesn't need to happen?

straleyb: If we are calling the method in fetch then yeah i dont think this is necessary.

app/authorities/qa/authorities/format.rb
straleyb: We should develop a strategy for this for sure. But right now, I think its ok if we keep this separate per authority. Since each authority is its own, im not sure that each return will be similar. Id much rather rely on something that works for now rather than worry about refactoring. Once we get all the CVs in for single and multi, maybe we can take a look at all the different formats and find something that works for all, but thats probably 20 or 30 different endpoints.

Issue #145, opened 2018-11-26
Build a Hyrax-friendly JP2 encoder using our logic


jechols:
### Descriptive summary

We want a way to use GraphicsMagick to encode from any image source into JP2.  We also want to make sure the encoded JP2 is as optimized as possible for RAIS to serve.  I've written up some general guidelines in the past about this: https://github.com/uoregon-libraries/rais-image-server/wiki/How-To-Encode-JP2s

### Expected behavior

- [x] When ingesting an image in any format (within reason), a JP2 is generated
  - We want to ensure our list of filetypes doesn't include those which aren't truly images, but which GraphicsMagick may support, such as PDFs or other binary files that happen to have images in them (see http://www.graphicsmagick.org/formats.html)
- [x] When a JP2 is generated, it has tiling at 1024x1024, and a sensible number of resolution levels based on its size
- [x] PDFs should go through a ... "fork" of this process, for lack of a better term.  They need one JP2 per page in order to make them easier to handle than our current "four JPGs per page" setup.
- [x] Perfect success would be defined as an implementation that works with Hydra really well, as is the case with https://github.com/samvera/hydra-derivatives/blob/master/lib/hydra/derivatives/processors/jpeg2k_image.rb
  - If it isn't easy to "register" our encoder in the Hyrax processor list, we should just have `lib/oregon_digital/derivatives/processors/jp2.rb` and just call it from `OregonDigital::FileSetDerivativesService`.  We can make it more like a Hyrax derivative processor once we have something working.

### Related work

#141, #78, #30

jechols:
Actually I think this may be doable just using the image processor Hyrax already has.  If not, I can almost definitely just subclass that.

jechols:
Maybe not - imagemagick doesn't have a flag for tiling... it uses a funky syntax on the output filename, e.g.: `convert wizard.png 'wizard.png[512x512]'`

jechols:
Nope, looks like we'll have a custom processor - I'm not seeing anything in the image processor that would let us specify the other flags we need, either.

jechols:
MiniMagick (and in fact imagemagick on the command line) cannot process large files - even the 430-megapixel image in my test set causes a crash.  I'm going to look at options.  This might be why we (UO) had to use GraphicsMagick.

jechols:
MiniMagic supports GraphicsMagick, but my GraphicsMagick install (in a test Ruby 2.5.1 image) doesn't handle JP2.  Investigations continue.

[...]

PR #363, opened 2019-02-15
Feature/jp2 encoding

jechols:
Addresses part of #145.  All image sources give us zoomable JP2 images without kakadu.  This breaks the PDF processor, however, because that's still trying to go through the Hyrax layer.  The Hyrax layer uses minimagick with assumptions that the base process will be ImageMagick.  We have to use GraphicsMagick because (at the moment) ImageMagick crashes on huge files.

<s>I request that, on merge, the commits stay as-is, not be squashed.  There is some value in the three minor commits not being smashed into the big bomb.</s>

This is flagged as "draft" because I still want to see if I can get PDFs to a state where they at least don't crash.  I'm not sure about getting them fully implemented with our custom logic (at least not as soon as I'd like), but I would like to get them to not crash.


jechols:
There's a weird issue converting gargantuan JP2s to BMPs I'm looking into.  Might be not worth holding this up, but I want to get an idea if I missed something or what.

PR #401, opened 2019-03-12
Force index strategy to be defined in metadata concerns

### LINKED ISSUE #397: Metadata not indexing as facetable in Solr
### Descriptive summary

Non-core metadata is only being indexed as `<property>_tesim` which relates to `Solrizer.solr_name(<property>, :stored_searchable)`. `Solrizer.solr_name(<property>, :facetable)` outputs `<property>_sim` and these values aren't found in the schema or documents once a work is saved.

This seems to be causing facets to not appear for most metadata fields on the search page, though may not be the only factor

### Expected behavior

Metadata that specifies to be indexed as `:stored_searchable` and `:facetable` should be indexed into Solr as `<property>_tesim` and `<property>_sim`

### Related work
Blocks work on #271

CGillen:
Fixes #397 by defaulting all metadata to the indexing methods defined in the metadata concerns. eg: https://github.com/OregonDigital/OD2/blob/master/app/models/concerns/oregon_digital/generic_metadata.rb#L22

### DIFF
--- a/app/models/concerns/oregon_digital/generic_metadata.rb
+++ b/app/models/concerns/oregon_digital/generic_metadata.rb
@@ -13,17 +13,29 @@ module OregonDigital
       property :label, predicate: ActiveFedora::RDF::Fcrepo::Model.downloadFilename, multiple: false
       property :relative_path, predicate: ::RDF::URI.new('http://scholarsphere.psu.edu/ns#relativePath'), multiple: false
       property :import_url, predicate: ::RDF::URI.new('http://scholarsphere.psu.edu/ns#importUrl'), multiple: false
-      property :resource_type, predicate: ::RDF::Vocab::DC.type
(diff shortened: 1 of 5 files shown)


app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -13,17 +13,29 @@ module GenericMetadata
...
+      end
 
       property :description, predicate: ::RDF::Vocab::DC.description do |index|
         index.as :stored_searchable
       end
 
-      property :rights_statement, predicate: ::RDF::Vocab::EDM.rights
-      property :identifier, predicate: ::RDF::Vocab::DC.identifier
+      property :rights_statement, predicate: ::RDF::Vocab::EDM.rights do |index|
+        index.as :stored_searchable
```
wickr: should be :facetable here as well

PR #438, opened 2019-03-22
Retry fetching linked data queue

### LINKED ISSUE #312: Create job that reruns a "fetch label" if a fetch label fails
### Descriptive summary

Per discussions on Thursday, when someone creates a work, and adds linked data to that work, there is a possibility that fetching the label for that url could fail. When that job fails we want to automagically fire off a new job to attempt to refetch that label. We want to continue to do this until the fetch works and the label is successfully pulled from the endpoint. 

After the job fails, we would like to send off a notification to that user notifying them that their data hasn't been fully saved (or better verbiage than that) so they know that their work doesn't have all the data that it would otherwise due to an endpoint being down. Once the job fetches properly, we should send off another notification telling them their work has been updated successfully. 

### Expected behavior

When a work has a uri in its form and is saved, it should go out and try to fetch the labels from the data source. If the fetch returns a failed code, it should enqueue a job to continue to fetch that label and notify the user that everything isnt updated in the work quite yet. With nice verbiage of course. When it is successful it should notify the user it was successfully updated. It should continuously run the job until it fetches successfully. 

- [x] User can save a work and have it fetch a label from an endpoint
- [x] Upon failure of fetching the label it should notify the user
- [x] Upon failure of fetching the label it should enqueue a job to refetch the label
- [x] It should continuously fire off jobs until the fetch is successful
- [x] It should notify the user and update the work upon successful fetching

CGillen:
Fixes #312 
Sidekiq automatically has an exponential backoff and requeues on exception: https://github.com/mperham/sidekiq/wiki/Error-Handling

~~Just needs user messaging, probably based on other Hyrax messaging. Like: https://github.com/samvera/hyrax/search?q=ImportUrlFailureService&unscoped_q=ImportUrlFailureService~~

When a fetch fails during a work save per URI:
- Enqueues a LinkedDataWorker job
- ~Adds stub label to BlazeGraph.~
- Emails depositor about failure

On run a LinkedDataWorker job will:
- ~Fetch and delete current graph~
- Attempt a new fetch from local, if successful exit early
- Attempt new fetch from remote
  - On failure, requeue for another attempt
  - On success, new graph is saved by TriplestoreAdapter and depositor is emailed about success
- On ~25~ 11 failures (approx ~21~ 2 days), depositor is emailed about exhausted retries and job is put in Dead queue. Can be run manually or deleted.


CGillen:
@revgum I'm not sure what's going on with CircleCI but the failing tests pass fine locally.
It's like these files aren't be added to the environment:
https://github.com/OregonDigital/OD2/pull/438/files#diff-408c5e81a87f9282ff808f7ad063af2f
https://github.com/OregonDigital/OD2/pull/438/files#diff-e2ec482443e906ae2b44d2d1f04833a2

Perhaps caching issues?

PR #439, opened 2019-03-25
require on autoloaded file

straleyb:


lib/oregon_digital/controlled_vocabularies.rb
```diff
@@ -1,12 +1,11 @@
 # frozen_string_literal: true
+Dir[File.dirname(__FILE__) + '/controlled_vocabularies/*.rb'].each {|file| require file }
```
CGillen: You should be able to add `config.eager_load_paths << Rails.root.join('lib')` to config/application.rb and get the same effect.

Then we should be able to get rid of everything that's `eager_autoload` in the lib directory

straleyb: I was talking with @revgum about this and he said he didnt really feel comfortable autoloading everything from lib. We decided on requiring everything in the CVs directory instead.

CGillen: Then you should be able to narrow it down to `'lib/oregon_digital/controlled_vocabularies'` and it should grab all sub-directories.

I'm thinking the hesitation comes from the ability to load any files an attacker could somehow write to disk in /lib. That seems to hold true for both of these implementations.

Either way works, so I won't block this one once it's passing, but I think we should consider the alternative if we come back around to a large amount of libs to include.

PR #446, opened 2019-03-26
Adds in new 404 errors processed through asset pipeline

straleyb:
Need to figure out how to get the pages copied over to the proper place when a deployment happens. Might need to happen on kubernetes? @revgum


config/initializers/assets.rb
```diff
@@ -8,9 +8,6 @@
 # Add additional assets to the asset load path.
 # Rails.application.config.assets.paths << Emoji.images_path
 # Add Yarn node_modules folder to the asset load path.
+Rails.application.config.assets.precompile += %w(404.html 500.html)
```
CGillen: According to https://guides.rubyonrails.org/asset_pipeline.html#search-paths and https://guides.rubyonrails.org/asset_pipeline.html#precompiling-assets these lines shouldn't be necessary, but I'll believe it if you tell me otherwise.

Dockerfile
```diff
@@ -51,6 +51,9 @@ RUN ./build/install_gems.sh
 # Add the application code
 ADD . /data
 
+RUN cp /data/app/assets/html/404.html.haml public/404.html
+RUN cp /data/app/assets/html/500.html.haml public/500.html
+
 RUN if [ "${RAILS_ENV}" = "production" ] || [ "${RAILS_ENV}" = "staging" ]; then \
   RAILS_ENV=${RAILS_ENV} SECRET_KEY_BASE=${SECRET_KEY_BASE} bundle exec rails assets:precompile; \
```
CGillen: Here's where assets are compiled for production. Add a `cp` or `rsync` here to move /app/assets/\*.html to /public/\*.html
Maybe add an else for copying .haml files when not on prod/staging, instead of above.

PR #450, opened 2019-03-28
Contact form message

CGillen:
Includes commits for just CC'ing the submitter and a custom message.
@petersec @vantuyls If CC'ing is good enough, it's cleaner code and I can revert to that commit. Otherwise checkout [`app/views/oregon_digital/contact_mailer/contact.html.erb`](https://github.com/OregonDigital/OD2/blob/924a86bea7472a581dfa5578288dd9059235cfe3/app/views/oregon_digital/contact_mailer/contact.html.erb) for the custom message.


straleyb:
Looks good to me. Once we get approval from @vantuyls and @petersec then ill smash the merge button.

petersec:
This looks fine to me as well.

Issue #443, opened 2019-03-25
Notification for workflows


vantuyls:
### Descriptive summary

When a work enters a workflow, notification should be sent to those assigned to the workflow step for the admin set. Message:

The work <Title> (<link>) has been uploaded and is ready for your review.

CGillen:
This behavior already happens with the message:
`<Title> (<ID as link>) was deposited by <Depositor email> and is awaiting approval`

The depositor and all users with the "approving" workflow role get this message as an in-app and email notification.
The original intent of this issue was to get the functionality working and change the contents later, so I will move this to QA. #451 shows how we can change the message.

vantuyls:
QA Pass - received message in system upon deposit

PR #451, opened 2019-03-28
updates notification message

straleyb:
Another update for notifications I discovered while figuring out the workflow for notifications via workflow states. This changes the notification to use the verbiage described in #443


straleyb:
Current message is good enough

PR #458, opened 2019-04-02
fixes ons

straleyb:


lib/oregon_digital/controlled_vocabularies/vocabularies/ons_base.rb
```diff
@@ -5,7 +5,7 @@ module ControlledVocabularies::Vocabularies
     # Receives information pulled from the endpoint and can parse and generate queries
     class OnsBase
       def self.label(data)
-        data.first['rdfs:label']['@value']
+        data['rdfs:label']['@value']
```
CGillen: This messes up other vocabularies. `Qa::Authorities::Scientific` needs to supply `response` as an array in `#find_term`. This will probably require changes in uBio and ITIS vocabularies `#label` methods.

PR #479, opened 2019-04-05
Work for daily email

straleyb:
Need to inspect when the work entered a queue, and if its the same day then send out email.


revgum:
Since the only thing that is different between the emails is the subject and the body text.. it might make sense to combine these into a single `NotificationMailer` and pass the subject and some piece of the body text when you fire off the mailer for each of the two contexts it'd be used for.

PR #484, opened 2019-04-09
refactor generation of singleton variable to a method

straleyb:


app/models/concerns/oregon_digital/document_metadata.rb
```diff
@@ -44,6 +43,11 @@ module DocumentMetadata
       property :table_of_contents, predicate: ::RDF::Vocab::DC.tableOfContents do |index|
         index.as :stored_searchable
       end
+      puts properties.keys
```
revgum: Remove your debugging `puts`

app/models/concerns/oregon_digital/document_metadata.rb
```diff
@@ -44,6 +43,11 @@ module DocumentMetadata
...
+
+      define_singleton_method :document_properties do
+        (properties.select { |_k, v| v.class_name.nil? }.keys - initial_properties)
```
revgum: The assumption here is that any property with a `class_name` must be a controlled vocabulary.. it would be better if this checked for a class_name and that the class name doesn't include `ControlledVocabularies`?

app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -601,52 +601,23 @@ module GenericMetadata
...
-      accepts_nested_attributes_for :transcriber, reject_if: id_blank, allow_destroy: true
-      accepts_nested_attributes_for :translator, reject_if: id_blank, allow_destroy: true
-      accepts_nested_attributes_for :cultural_context, reject_if: id_blank, allow_destroy: true
-      accepts_nested_attributes_for :style_or_period, reject_if: id_blank, allow_destroy: true
-      accepts_nested_attributes_for :form_of_work, reject_if: id_blank, allow_destroy: true
-      accepts_nested_attributes_for :institution, reject_if: id_blank, allow_destroy: true
-      accepts_nested_attributes_for :workType, reject_if: id_blank, allow_destroy: true
+
+      # Sets controlled values
+      self.controlled_properties = properties.reject { |_k, v| v.class_name.nil? }.keys.map(&:to_sym)
```
revgum: The assumption here is that any property with a `class_name` must be a controlled vocabulary.. it would be better if this checked for a class_name and that the class name includes `ControlledVocabularies`?

app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -601,52 +601,23 @@ module GenericMetadata
...
+        accepts_nested_attributes_for prop, reject_if: id_blank, allow_destroy: true
+      end
+
+      # defines a method for Generic to be able to grab a list of properties
+      define_singleton_method :controlled_props do
+        controlled_properties.each_with_object([]) { |prop, array| array << "#{prop}_label" }.freeze
+      end
+
+      define_singleton_method :generic_properties do
+        (properties.select { |_k, v| v.class_name.nil? }.keys - initial_properties)
```
revgum: The assumption here is that any property with a `class_name` must be a controlled vocabulary.. it would be better if this checked for a class_name and that the class name doesn't include `ControlledVocabularies`?

app/models/concerns/oregon_digital/image_metadata.rb
```diff
@@ -37,6 +36,10 @@ module ImageMetadata
       end
 
       property :width, predicate: ::RDF::Vocab::EXIF.width, multiple: false
+
+      define_singleton_method :image_properties do
+        (properties.select { |_k, v| v.class_name.nil? }.keys - initial_properties)
```
revgum: The assumption here is that any property with a `class_name` must be a controlled vocabulary.. it would be better if this checked for a class_name and that the class name doesn't include `ControlledVocabularies`?

app/models/concerns/oregon_digital/video_metadata.rb
```diff
@@ -7,12 +7,13 @@ module VideoMetadata
...
 
-    PROPERTIES = %w[height width].freeze
-
     included do
+      initial_properties = properties.keys
       property :height, predicate: ::RDF::Vocab::EXIF.height, multiple: false
-
       property :width, predicate: ::RDF::Vocab::EXIF.width, multiple: false
+      define_singleton_method :video_properties do
+        (properties.select { |_k, v| v.class_name.nil? }.keys - initial_properties)
```
revgum: The assumption here is that any property with a `class_name` must be a controlled vocabulary.. it would be better if this checked for a class_name and that the class name doesn't include `ControlledVocabularies`?

app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -601,52 +601,23 @@ module GenericMetadata
...
+      # Sets controlled values
+      self.controlled_properties = properties.reject { |_k, v| v.class_name.nil? }.keys.map(&:to_sym)
+
+      # Allows for the controlled properties to accept nested data
+      controlled_properties.each do |prop|
+        accepts_nested_attributes_for prop, reject_if: id_blank, allow_destroy: true
+      end
+
+      # defines a method for Generic to be able to grab a list of properties
+      define_singleton_method :controlled_props do
```
revgum: this method should be named `controlled_properties` so that throughout the code it looks similar to how the `*_properties` methods look.

app/models/concerns/oregon_digital/document_metadata.rb
straleyb: Thats a good question. The assumption was that the only properties with a class_name are the ones that are controlled. 
https://github.com/OregonDigital/OD2/blob/master/app/models/concerns/oregon_digital/generic_metadata.rb#L491 This is where the property gets a class_name associated with it. There SHOULD be no other properties that get a class_name associated with it. But we could inspect deeper and say `select all values with a classname that doesnt have ControlledVocabularies` so that if there are any other classes that get associated with a property, it wont skip those.

app/models/concerns/oregon_digital/generic_metadata.rb
straleyb: As stated above, the class_name gets populated with the classes that are associated with a property. Which SHOULD only be controlled vocabularies, but it could change.

app/models/concerns/oregon_digital/generic_metadata.rb
straleyb: I was afraid that would be a name collision and could effect other parts of the system. I can do more thorough testing however to see if thats actually the case.

app/forms/hyrax/generic_form.rb
```diff
@@ -26,7 +26,7 @@ def secondary_terms
 
     def self.build_permitted_params
       params = super
-      OregonDigital::GenericMetadata::CONTROLLED.each do |prop|
+      Generic.generic_properties.each do |prop|
```
CGillen: `Generic.controlled_properties`?

app/models/concerns/oregon_digital/generic_metadata.rb
straleyb: Stack level too deep when defining this method as #controlled_properties

revgum: Is this because `controlled_properties` is already defined as a class_attribute in lines 603-611?

CGillen: That or because it calls `controlled_properties` in the body and it's a namespacing issue.

This might be more appropriately called `controlled_labels` anyways, since these are all `<prop>_label` values.
This might also raise the question if these should all be defined as `self.*_properties` for consistency?

PR #485, opened 2019-04-09
Feature/pdf derivatives

### LINKED ISSUE #374: Create PDF derivatives
### Descriptive summary

Building on the JP2 derivatives for images, we want to build JP2s for PDFs: one JP2 per PDF page.

### Expected behavior

PDFs should go through a processor that's very similar to what images go through, but instead of a single JP2 being generated, we want one per page.  This is a lot of derivatives, but it will result in a much smaller footprint than OD1's setup which generates four separate JPGs per PDF page.

PDFs still need a single thumbnail for search display.  This should probably just be based on the first page since that's what's happening right now.

### Related work

#141, #78, #30, #145

jechols:
Closes #374

This generates the PDF thumbnail as before, fixes the lack of PDF OCR (I ripped that out by mistake previously), and generates one JP2 per PDF page.  Displaying those pages... that's the 80% I'm still working on (see #493).

### DIFF
--- a/app/services/hyrax/derivative_path.rb
+++ b/app/services/hyrax/derivative_path.rb
@@ -1,0 +1,67 @@
+module Hyrax
+  class DerivativePath
+    attr_reader :id, :destination_name
+
+    class << self
+      # Path on file system where derivative file is stored
+      # @param [ActiveFedora::Base or String] object either the AF object or its id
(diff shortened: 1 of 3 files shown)


jechols:
ack, forgot tests

PR #517, opened 2019-04-19
Add style guide to footer.

### LINKED ISSUE #515: Add link to style guide.
### Descriptive summary

As a developer, I want to know the latest look and feel in a central place so I can contribute more.

### Expected behavior

I'll add a link in the OD2 footer to the "living style guide" alongside versioning and other more administrative references.

### Related work

mccallum:
Linked style guide in footer.... might eventually become admin only like other info there.

Closes #515


mccallum:
@revgum Yes, but eventually, it'll want to be tied to being logged-in role so as not to detract from end user/researcher. PMPOSO may disagree but I'm in favor of loud, proud, open for living style guides -- as many organizations are (a la Material, Lightning, US Gov, etc).  And yeah, it should be easy to switch ownership if now is a good time for that.

mccallum:
@revgum I can code it to only show when loggedin.

revgum:
Cool, yea lets change ownership.. and you should be able to surround your link with the following to essentially make sure a user is an admin:
```
<% if can? :read, :admin_dashboard %>
  <a href.....
<% end %>
```

vantuyls:
deleted per @straleyb recommendation

PR #540, opened 2019-05-21
Add Blacklight facets to collections search page

### LINKED ISSUE #532: Collections landing/search page should display facets if configured to do so
### Descriptive summary

After the collection landing/search page accepts facets, facets should display on the page, so people can... facet.

### Expected behavior

Facets display on collection landing/search page. Facets correctly facet search, per #531. Display and functionality should be parallel to the works search page.

### Related work

Depends on #531 
Part of #528 

### Accessibility Concerns

Facets added to display will add new HTML and possibly JS functionality. It should be the same as the works search pages, and ideally generated the same way so accessibility fixes applied to one affect the other.

CGillen:
Fixes #532.
Display facets exactly as catalog search page does, from Blacklight.
`_facets.html.erb` isn't necessary yet, but will require changes once we can get a list of facets to display for the collection.

Preview of collections search page:
image


straleyb:
If these are overridden, can you add OVERRIDE to them? And comment where the overrides are?

PR #549, opened 2019-05-30
Add configurable facets to collections edit form

### LINKED ISSUE #547: Hook up collection code to new collection type facet_configurable code
### Descriptive summary

With the collection type code set up for a collection type to know if its facets are configurable, we need to set up the ability for a collection to render facets in the proper place. This has to check whether or not a specific collection's collection type allows for configurable facets.

A collection should respond to `collection.facet_configurable?` and return a true or false statement.

CGillen:
Fixes #547 
This PR creates a Facet object that is used to store configuration for facets for each collection. It also adds the configurable facet tab to the collections edit form and hooks it up to that facet object. 
Draggable ordering is accessible with tab-focus and (left/up)/(right/down) keyboard navigation.

Preview:
image

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -18,7 +18,12 @@ class Collection < ActiveFedora::Base
     properties = properties.merge(Image.properties)
     properties = properties.merge(Video.properties)
     properties.select { |_k, prop| prop.behaviors && prop.behaviors.include?(:facetable) }.each do |_k, prop|
-      facet = Facet.new({label: prop.term.to_s, solr_name: Solrizer.solr_name(prop.term, :facetable), collection_id: self.id, property_name: prop.term})
+      facet = Facet.new({
+        label: I18n.translate("simple_form.labels.defaults.#{prop.term}"),
+        solr_name: Solrizer.solr_name(prop.term, :facetable),
+        collection_id: self.id,
+        property_name: prop.term
+      })
       facets += [facet] if facets.select { |f| f.property_name == facet.property_name }.empty? && facet.save
     end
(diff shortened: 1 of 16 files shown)


straleyb:
This is looking pretty good. Just some simple tests to cover some stuff a bit more would do just fine.

app/models/collection.rb
```diff
@@ -2,8 +2,47 @@
...
 # Sets the behaviors and other data for a collection
 class Collection < ActiveFedora::Base
+  after_destroy :destroy_facets
+
   include ::Hyrax::CollectionBehavior
   # You can replace these metadata if they're not suitable
   include Hyrax::BasicMetadata
   self.indexer = Hyrax::CollectionWithBasicMetadataIndexer
+
+  delegate(:facet_configurable?, to: :collection_type)
```
straleyb: this should be fine as `delegate :facet_configurable?, to: :collection_type`

CGillen:
This still has accessibility concerns with keyboard use and screen reader instructions.
https://salesforce-ux.github.io/dnd-a11y-patterns/#/sortA?_k=z62gvk offers a better example of key bindings and using an aria-live element to provide instructions and feedback about item movement. I'll open a new ticket about this and link this PR, comment, and example.

CGillen:
3be1c52 included fixes for #551. NVDA throws a fit with it, and completely mucks with how the browser natively handles things. It works on chrome+FF by themselves though. I'm not taking NVDA as gospel, and don't think we should code to get a specific screen reader working, but rather focus on what should be correct ARIA/HTML

PR #562, opened 2019-06-05
create colls task

lsat12357:
fixes OregonDigital/hyrax-migrator#119


lsat12357:
based on a conversation with @wickr: the Metadata folks would like osu-scarc to be an admin set AND a coll, so he suggested off the top of his head that the admin set id be osu-scarc-admin

lib/tasks/oregon_digital/create_collections.rake
```diff
@@ -0,0 +1,38 @@
...
+    rescue StandardError => e
+      puts "Unable to create collection #{id}"
+      puts "Error: #{e.message}"
+    end
+  end
+end
+
+def collection_exists?(id)
+  c = Collection.where(id: id).first
+  c.present?
```
luisgreg99: It looks like `Collection` supports `exists?`, so we can also do `Collection.exists?(id)` to simplify.

PR #563, opened 2019-06-05
bring over admin_set crosswalk file to OD2

lsat12357:
This brings over admin set crosswalk file from the migrator to live in OD2 configs with the other .yml files so the service can find it. Modified the id of osu-scarc (see comment on #562)


lsat12357:
oop, I guess the admin set ids and therefore the crosswalk are being updated iminently? so maybe sit on this?

revgum:
@lsat12357 ping with an update on the status/mergability?

Issue #520, opened 2019-05-17
Translate design for OD2 home page


straleyb:
### Descriptive summary

We need to take the page @mccallum drew up and translate it into code. The page can be found here: https://sketch.cloud/s/mEz3z/bgDKb0V

If we can get this to a point thats "close enough" with some of the things not completely working yet (I.E the for educators and for researchers buttons) Then that would be considered "close enough"

CGillen:
Round 1 was #566 
Some more things to work on:
- CSS for images, color changes, etc
- Library Badges to left of intro
- Brand image at top left
- Hero image/background

raybrarian:
Closing this one - the work (in its current, PO/SM-approved simplified form) has been broken into individual tickets based on the new wireframes.

PR #566, opened 2019-06-07
Homepage Design

CGillen:
First pass at #520 
image

Educator and Researcher buttons don't work, collections and items do.


CGillen:
image
image

More tweaks and the masthead. 
A LOT of other pages are effected color-wise, but I don't think anything is breaking. Some pages might look weird.

revgum:
Looks like a great start!

PR #657, opened 2019-07-30
update homepage with how do i use section

### LINKED ISSUE #601: Homepage: Add new 'How Do I Use Oregon Digital?' section
### Descriptive summary

Similar to the section on http://dp.la and instead of the For Educators and For Researchers buttons we had, add a new section called "How Do I Use Oregon Digital? with 4 image block sections:

4 block links to information pages, styled similar to DPLA homepage, placeholder icons with text for now:
* Students
* Teachers
* Researchers
* Lifelong Learners

We don't have icons designed or selected yet, one option may be Noun Project icons, inquiries on using these have been made.

This will go below the new Showcase section.

Wireframe mockup: https://www.lucidchart.com/documents/edit/9814ebf7-6286-406c-946b-8fa69ff92cbb/0

(Information compiled by POSM group from updated wireframes)

### Expected behavior

New section shows links to different information pages relevant to our audiences.

straleyb:
Adds new how to I use section. Need to make it look nicer, but this gets it into place.

Fixes #601


straleyb:
Screen Shot 2019-07-30 at 1 21 05 PM

We can refine the look but this gets it most the way there. Clickable to a placeholder page.

app/views/hyrax/homepage/_how_to_use.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="col-md-2"></div>
+<a href="<%= Hyrax::Engine.routes.url_helpers.help_path %>">
+  <div class="col-xs-4 col-sm-2 homepage-box">
```
CGillen: Is there a reason you went with smaller boxes in the center? I thought this was supposed to be full-width, so it should be 4 `col-sm-3` divs. Then I would think we'd drop to 2 wide on mobile, so `col-xs-6`. I mostly just want to get this consistent, so let's decide on one and I can fix my PR either way.

app/views/hyrax/homepage/_how_to_use.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="col-md-2"></div>
```
CGillen: Instead of a dummy div, try `col-md-offset-2` on the first element below
or `col-sm-offset-2` in this case I guess.

app/views/hyrax/homepage/_how_to_use.html.erb
straleyb: We went with the more centered look.

app/views/hyrax/homepage/_how_to_use.html.erb
straleyb: col-sm-offset-2 used successfully

app/views/hyrax/homepage/_how_to_use.html.erb
CGillen: Resolved with Slack convo. Targeting this design https://dp.la/ as it was the base reference.
Can change to full-width later if desired.

PR #661, opened 2019-07-31
change to links

straleyb:


app/views/_controls.html.erb
```diff
@@ -9,10 +9,10 @@
   <div class="row">
     <div class="col-md-4"></div>
```
CGillen: No dummy divs please

app/views/_controls.html.erb
```diff
@@ -9,10 +9,10 @@
...
     <div class="col-md-1">
```
CGillen: Try `col-xs-push-1 col-sm-7 col-sm-push-3 col-md-5 col-md-push-4` (to match the search box) and move both links into the save div.
This should fix up the extra scrolling and keep the links inline.

PR #665, opened 2019-08-01
Upgrade beeline to 1.0.0

decimalator:


Gemfile
```diff
@@ -41,16 +41,16 @@ gem 'omniauth-cas'
 gem 'omniauth-saml'
 
 # Honeycomb
-gem 'sequel'
-gem 'honeycomb-beeline'
-gem 'rack-honeycomb', '~> 0.5.0'
+#gem 'sequel'
```
CGillen: If we're cutting gems, lets try not to just comment them out.
We discussed not wanting to see code behind comments as much as possible during standups, 8/5.

Gemfile
```diff
@@ -41,16 +41,16 @@ gem 'omniauth-cas'
...
+#gem 'rack-honeycomb', '~> 0.5.0'
+gem 'honeycomb-beeline', '>= 1.0.0'
 
 # Recaptcha
 gem "recaptcha", require: "recaptcha/rails"
 
-group :production do
-  gem 'clamav', git: 'https://github.com/eagleas/clamav', branch: 'master'
-end
+#group :production do
```
CGillen: If we're cutting clamav, it would be nice to see it as a separate issue with some reasoning behind it for documentation reasons.

config/initializers/honeycomb.rb
```diff
@@ -0,0 +1,14 @@
+Honeycomb.configure do |config|
+  config.write_key = ENV.fetch('HONEYCOMB_WRITEKEY', 'd5529d94d4e989bb1fc1b1cfc559ebba')
```
CGillen: Is this a secret we should be protecting?

config/initializers/honeycomb.rb
```diff
@@ -0,0 +1,14 @@
...
+  config.dataset = ENV.fetch('HONEYCOMB_DATASET', 'od2-rails-staging')
```
CGillen: If this is environment, is it vaguely leaking our honeycomb config? Should the default focus for local development? This might be a non-issue.

config/initializers/honeycomb.rb
decimalator: ah yup good catch

config/initializers/honeycomb.rb
decimalator: yeah we can make the defaults be something different as long as it can pass CCI tests, or we can not configure it for test

Gemfile
decimalator: we can cut clamav in a separate PR but clamav is currently broken and fails builds

Gemfile
decimalator: :+1:

straleyb:
Moved to #677

PR #676, opened 2019-08-06
Works can generate temporary CSV metadata exports

### LINKED ISSUE #666: Export metadata as CSV
### Descriptive summary
As part of #644, a work should be able to export its metadata to a temporary CSV file. The first row should be headers of the human readable names for each metadata entry. 
The data should be one row of metadata.
Edit from conversation with Kate:
Multivalue metadata should be a quotes escaped list of values, separated by `|` (bars/pipes). 
~Labels for CVs should not be provided, just URIs.~ Both labels and URIs should be present for CVs.

Use the standard CSV library to avoid issues with the CSV spec: https://ruby-doc.org/stdlib-2.6.1/libdoc/csv/rdoc/CSV.html
And if you still think the format looks off, check the RFC: https://tools.ietf.org/html/rfc4180

### Expected behavior
A work can provide a file pointer to a temporary CSV metadata export. The CSV file is formatted to RFC4180.

For now, ignore file clean up and we'll deal with it in a separate issue (#667).

### Related work
Child of #644 

### Accessibility Concerns
None

CGillen:
Fixes #666 
CSVs are saved to `<rails>/tmp/works/metadata`
Values are separated by `|`
Controlled vocab is exported as `label [uri]`

Example:
```
has_model,create_date,modified_date,depositor,title,date_uploaded,date_modified,resource_type,rights_statement,identifier,license,institution,local_collection_name
Generic,2019-07-22T22:49:12+00:00,2019-08-01T19:45:59+00:00,admin@example.org,Generic Title,2019-07-22T22:49:10+00:00,2019-08-01T19:45:59+00:00,http://purl.org/dc/dcmitype/Collection,http://rightsstatements.org/vocab/InC/1.0/,"test|third, with comma|second identifier as free text, adding comma in second",http://creativecommons.org/publicdomain/zero/1.0/,Oregon State University [http://dbpedia.org/resource/Oregon_State_University]|University of Oregon [http://dbpedia.org/resource/University_of_Oregon],"Mechanical Engineering Department Records, 1894-1971 (RG 213) [http://opaquenamespace.org/ns/localCollectionName/rg_213]"
```


straleyb:
Can we get some tests to cover these new methods?

app/models/concerns/oregon_digital/work_behavior.rb
```diff
@@ -42,5 +60,42 @@ def resolve_oembed_errors
...
+        props[label] = values.map(&:to_s).join('|')
+      end
+      props
+    end
+
+    # Convert work controlled property value to '<label> [<uri>]' format
+    def controlled_property_to_csv_value(prop)
+      prop.fetch
+      prop = prop.solrize[1][:label].split('$')
+      prop[1] = '[' + prop[1] + ']'
```
straleyb: why not `"[#{prop[1]}]"`. I prefer string interpolation instead.

PR #677, opened 2019-08-06
initial ruby bump

### LINKED ISSUE #659: Update OD2 to Ruby 2.5.5
### Descriptive summary

We need to bring up the ruby version to ruby 2.5.5

### LINKED ISSUE #678: Remove antivirus detection (clamav)
### Descriptive summary
Reversing #325 since the clamav gem is ancient and breaks with the switch from Debian Stretch to Buster. Our file upload process is mediated by a small, trusted team, and upload is non-anonymous. We're not as protected as having an AV, but we will be fairly mitigated.

### Expected behavior
ClamAV is no longer installed.

### Related work
#325, #619 

### Accessibility Concerns
None

straleyb:
Fixes #659 
Fixes #678 

Need to figure out what to do with the secret keys and make honeycomb not run for tests


straleyb:
We removed CLAMAV because the CLAMAV gem is not compatible with ruby version 2.5.5

Dockerfile
```diff
@@ -13,17 +13,8 @@ RUN gem install bundler
...
-
-# install clamav for antivirus
-# fetch clamav local database
-RUN apt-get install -y clamav-freshclam clamav-daemon libclamav-dev
-RUN mkdir -p /var/lib/clamav && \
-  wget -O /var/lib/clamav/main.cvd http://database.clamav.net/main.cvd && \
-  wget -O /var/lib/clamav/daily.cvd http://database.clamav.net/daily.cvd && \
-  wget -O /var/lib/clamav/bytecode.cvd http://database.clamav.net/bytecode.cvd && \
-  chown clamav:clamav /var/lib/clamav/*.cvd
+  build-essential libpq-dev libreoffice imagemagick graphicsmagick unzip ghostscript vim postgresql-11\
```
CGillen: Move back postgresql-11

docker-compose.yml
```diff
@@ -259,10 +259,10 @@ services:
   ##
   # Postgres database server
   db: &db
-    image: postgres:10-alpine
+    image: postgres:11-alpine
```
CGillen: Walk this back

docker-compose.yml
```diff
@@ -259,10 +259,10 @@ services:
...
     networks:
       internal:
-  db-dev:
+  db-dev: &db-dev
```
CGillen: Haven't checked this change, is it necessary? Don't know enough about this syntax to call it

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
CGillen: Get this yarn out of here!

.rubocop_fixme.yml
```diff
@@ -71,3 +71,11 @@ RSpec/MultipleExpectations:
 RSpec/NestedGroups:
   Exclude:
     - 'spec/controllers/hyrax/contact_form_controller_spec.rb'
+
+RSpec/LeakyConstantDeclaration:
+  Exclude:
+    - 'spec/views/hyrax/admin/collection_types/_form_settings.html.erb_spec.rb'
```
CGillen: These could be moved into a let block to fix this

.rubocop_fixme.yml
```diff
@@ -71,3 +71,11 @@ RSpec/MultipleExpectations:
...
+
+RSpec/EmptyLineAfterHook:
+  Exclude:
+    - 'spec/models/generic_spec.rb'
```
CGillen: Whitespace isn't bad

docker-compose.yml
straleyb: Id like to leave this here. It helps to be able to spin up containers alone to check on their state.

PR #692, opened 2019-08-08
Catalog Controller Untangling

### LINKED ISSUE #629: Untangle search functionality from show functionality in catalog controller
### Descriptive summary

The way it stands, when something is configured as `:stored_searchable` it both tries to search and add it to the show page. We need to find a neat way of being able to decide whether something wants to just be displayed, just be searched, just be advanced_searched, be displayed and advanced_searched, be displayed and be simple_searched, and every other permutation of those. 

This means looking at the MAP and seeing which fields want to be displayed but not searched, and all the other permutations and making sure the searching is completely aligned with the MAP. 

This will have to be cross checked with the MAP by a metadata person for QA.

straleyb:
Fixes #629 

Need to update code to work with `showable: false`

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -160,20 +160,20 @@ class CatalogController < ApplicationController
     #
     #
     ##########################################################################
-    Generic.controlled_property_labels.each do |prop|
-      label = prop.gsub('_label', '')
+    Generic.controlled_property_labels.each do |label|
+      prop = label.gsub('_label', '')
 
       # Skip if this property isn't indexed
-      next if Generic.properties[label].behaviors.nil?
+      next if Generic.properties[prop].behaviors.nil?
 
       # Add property as searchable all fields box and individually
-      if Generic.properties[label].behaviors.include?(:stored_searchable)
-        config.add_show_field solr_name(prop, :stored_searchable) if Generic.properties[label]['showable'] || Generic.properties[label]['showable'].nil?
+      if Generic.properties[prop].behaviors.include?(:stored_searchable)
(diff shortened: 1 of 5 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -92,9 +92,9 @@ def self.modified_field
 
       # Add property as searchable all fields box and individually
       if Document.properties[prop].behaviors.include?(:stored_searchable)
-        config.add_show_field solr_name(prop, :stored_searchable) unless Document.properties[prop]['advance_search']
+        config.add_show_field solr_name(prop, :stored_searchable) if Document.properties[prop]['showable'] || Document.properties[prop]['showable'].nil?
 
-        if Document.properties[prop]['advance_search'] || Document.properties[prop]['advance_search'].nil?
+        if Document.properties[prop]['basic_searchable'] || Document.properties[prop]['basic_searchable'].nil?
           config.add_search_field(prop) do |field|
```
CGillen: `config.add_search_field` does not control whether metadata is searchbox searchable, only if it gets a dropdown/specific field search (useful for advanced search I suppose).
Documentation on `add_search_field`: http://jessiekeck.com/customizing-blacklight/search_fields/

We need to mess with this block to only add metadata that is `basic_searchable` to the `all_fields` search field: https://github.com/OregonDigital/OD2/blob/master/app/controllers/catalog_controller.rb#L210-L217

CGillen: We also should look at what OTHER fields are added from `config.show_fields.values` before we rip out that statement and replace it with loops/joins and filters from `Document.properties` and others.

PR #708, opened 2019-08-12
Fix deep indexing for single value CVs

### LINKED ISSUE #707: Single value Repository does not allow works to be saved
### Descriptive summary
#649 Has caused the work edit form to break.
Hyrax does not know how to index deep metadata for fields that are single value. Specifically it assumes deep metadata is multivalue when it goes to fetch external labels: https://github.com/samvera/hyrax/blob/v2.5.0/app/indexers/hyrax/deep_indexing_service.rb#L32

### Expected behavior
Work edit form allows works to create/save.

### Related work
#649 
#462 

### Accessibility Concerns
None

CGillen:
Fixes #707

### DIFF
--- a/app/indexers/oregon_digital/deep_indexing_service.rb
+++ b/app/indexers/oregon_digital/deep_indexing_service.rb
@@ -12,6 +12,28 @@ module OregonDigital
 
     private
 
+    # OVERRIDEN FROM HYRAX to allow for single value CVs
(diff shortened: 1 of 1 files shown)


straleyb:
Went with changing the field back to multiple and adhering to the way the form works already. Migrator fix is to cast the single valued CV to an array.

PR #715, opened 2019-08-13
update css and style

straleyb:


app/assets/stylesheets/oregon_digital/_footer.scss
```diff
@@ -23,9 +23,9 @@
 .footer-link-text {
   color: white;
   text-decoration: underline;
-  :hover {
```
CGillen: `&:hover` should do it and more sassy

CGillen: And add `&:focus`

PR #722, opened 2019-08-14
Search List view: Field rendering pipelines

### LINKED ISSUE #683: Search Results: List View: Metadata fields update
### Descriptive summary

On Search Results, List View, change the metadata fields that are shown (below Title):

  - Creator: show up to 3 creators, display 'and others' when more than 3
  - Date
  - Description: Truncate to 20 words, with ellipsis
  - Resource Type

Wireframe mockup, List View: https://www.lucidchart.com/documents/edit/517560ad-afba-41c5-a2c5-d3ff2b16f2bf/0

(Information compiled by POSM group from updated wireframes)

### Expected behavior

Metadata fields shown in search results are changed to Creator, Date, Description, and Resource Type.

CGillen:
Fixes #683 

Depends on #719 and will require conflict resolution.

image


app/controllers/catalog_controller.rb
```diff
@@ -74,11 +74,11 @@ def self.modified_field
...
     config.add_index_field solr_name('license_label', :stored_searchable), label: 'License', link_to_search: solr_name('license_label', :facetable)
-    config.add_index_field solr_name('type_label', :stored_searchable), label: 'Resource Type', link_to_search: solr_name('type_label', :facetable)
     config.add_index_field solr_name('file_format', :stored_searchable), link_to_search: solr_name('file_format', :facetable)
     config.add_index_field solr_name('identifier', :stored_searchable), helper_method: :index_field_link, field_name: 'identifier'
     config.add_index_field solr_name('embargo_release_date', :stored_sortable, type: :date), label: 'Embargo release date', helper_method: :human_readable_date
     config.add_index_field solr_name('lease_expiration_date', :stored_sortable, type: :date), label: 'Lease expiration date', helper_method: :human_readable_date
+    config.add_index_field solr_name('type_label', :stored_searchable), label: 'Resource Type', link_to_search: solr_name('type_label', :facetable), if: false
```
straleyb: `if: false` what would be evaluating to false in this context?

app/presenters/blacklight/rendering/max_length.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Blacklight
```
straleyb: Is this an Override?

app/presenters/blacklight/rendering/max_values.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Blacklight
```
straleyb: Is this an override as well?

app/controllers/catalog_controller.rb
CGillen: Choosing whether to display the field in the index/search page, similar to title above. We add it as an index field here to easily render it later (with the icon at the bottom).

app/views/catalog/_index_list_default.html.erb
```diff
@@ -1,28 +1,30 @@
...
-  <div class="col-md-3">
-    <div class="collection-counts-wrapper">
-      <div class="collection-counts-item">
-        <span><%= collection_presenter.total_viewable_collections %></span>Collections
-      </div>
-      <div class="collection-counts-item">
-        <span><%= collection_presenter.total_viewable_works %></span>Works
-      </div>
+<% if document.collection? %>
+<% collection_presenter = Hyrax::CollectionPresenter.new(document, current_ability) %>
```
straleyb: i think you missed some tabs here.

app/presenters/blacklight/rendering/max_length.rb
CGillen: Nope, new class. Could be namespaced to OregonDigital, but it's completely compatible with BL. Same with other max_values below.

app/views/catalog/_index_list_type_icon.html.erb
```diff
@@ -0,0 +1,19 @@
+<% case type.downcase %>
```
straleyb: I have a proposition about this. I think we can simplify this view. I wonder if we can add a method to the document.
`<%= render 'index_list_type_icon', type: document.type_label.first %>` could be changed to `<%= render 'index_list_type_icon', type: document.type_to_class %>` where #type_to_class (name pending) could translate the type to its font awesome class. Then this view could just have `<i class="fa #{fa_class}" aria-hidden="true"></i>` in it. I just feel like having a huge switch case in the view is messy.

CGillen: For sure. It was down to weighing a switch on a model that has almost no knowledge about FA and minimum custom methods or this view. Both messy in their own right.
I guess a hash could accomplish the same thing.

Check out 733b2cf and see if that's the direction you're thinking

PR #742, opened 2019-08-20
js for checking boxes and checkbox positioning

straleyb:
Screen Shot 2019-08-20 at 1 44 27 PM
Screen Shot 2019-08-20 at 1 44 11 PM

### DIFF
--- a/app/assets/javascripts/oregon_digital/add_to_my_shelf.js
+++ b/app/assets/javascripts/oregon_digital/add_to_my_shelf.js
@@ -1,6 +1,6 @@
 $( document ).ready(function() {
   $('#Select_all_on_page').click(function() {
-    var select_object = this.checked;
-    $('.my_shelf_result_checkbox').each(function () { this.checked = select_object; });
+    var all_on_page_state = this.checked;
+    $('.my_shelf_result_checkbox').each(function () { this.checked = all_on_page_state; });
   });
 });
--- a/app/assets/javascripts/oregon_digital/add_to_my_shelf.js
+++ b/app/assets/javascripts/oregon_digital/add_to_my_shelf.js
@@ -1,5 +1,6 @@
 $( document ).ready(function() {
   $('#Select_all_on_page').click(function() {
-    $('.my_shelf_result_checkbox').each(function () { this.checked = !this.checked; });
+    var select_object = this.checked;
+    $('.my_shelf_result_checkbox').each(function () { this.checked = select_object; });
   });
 });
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/oregon_digital/add_to_my_shelf.js
```diff
@@ -0,0 +1,5 @@
+$( document ).ready(function() {
+  $('#Select_all_on_page').click(function() {
+    $('.my_shelf_result_checkbox').each(function () { this.checked = !this.checked; });
```
CGillen: This doesn't seem to work as expected for a mix of already checked and unchecked boxes

app/views/_controls.html.erb
```diff
@@ -5,9 +5,9 @@
         <%= render partial: 'catalog/search_form' %>
       </div>
       <div class="col-md-offset-10">
-        <%= button_to "Add results to My Shelf", root_path, class: 'btn btn-primary my-shelf-button' %>
-        <%= label_tag "Add All on page to My Shelf" do %>
-          <% raw("#{check_box_tag 'Add All on page to My Shelf'} Add All on page to My Shelf") %>
+        <%= button_to "Add to My Shelf", root_path, class: 'btn btn-primary my-shelf-button' %>
+        <%= label_tag "Add to My Shelf" do %>
```
CGillen: The label is for the checkbox, not the button. `<%= label_tag "" do %>`

app/assets/stylesheets/oregon_digital/_search.scss
```diff
@@ -4,4 +4,9 @@
 
 .my-shelf-button {
   margin-top: 15px;
+}
+
+.shift-right {
```
CGillen: This is problematic with long titles
image

Maybe we can work it into the resource-type div and simplify it to a `class="pull-right"`?

straleyb: I tried pull right but it didnt seem to operate as expected.

app/assets/javascripts/oregon_digital/add_to_my_shelf.js
straleyb: Yeah so if some are checked, and some arent, we check all the unchecked ones and leave the others. If all are checked, then uncheck them. That sound about right?

CGillen: Sounds right, but complicated. I think we get the same outcome by setting all .my_shelf_result_checkbox checked values to #Select_all_on_page checked value

straleyb: Yeah I see what you are saying. we should be able to push the #Select_all_on_page object through then check its state. 👍

app/assets/stylesheets/oregon_digital/_search.scss
CGillen: pushed 7b547ab as a fix for works with long names

image

PR #776, opened 2019-09-26
Add show ability

### LINKED ISSUE #756: View an item
### Descriptive summary

Exact same as #755

CGillen:
Fixes #756 
~Does not provide any restriction as is, still needs checks for access~
Can easily be made to work with visibility from #775


CGillen:
QA is slightly different than the other abilities.
The term 'view' means to reach the work show page, NOT see the existence of. It is possible to not be able to 'view' a work and to know that it exists (ie. it shows in search results).

- [ ] Unauthenticated users can NOT view works restricted to Institution
- [ ] Authenticated users not in BOTH a UO and OSU related role can NOT view works restricted to Institution
- [ ] Authenticated users IN BOTH a UO and OSU related role CAN view works restricted to Institution

Final expected behaviour depends on #775.

PR #784, opened 2019-09-27
work on setting up initial ability for review

straleyb:
This is the basis for the work that needs to be done. Collection Managers need to be able to review works in their collections. We need to be able to see what collection a work is a part of and see if the person viewing that work is a collection manager for that work. This is merely the beginning work into discovering how to pass variables into an ability.

### DIFF
--- a/app/models/concerns/oregon_digital/ability/work_review_ability.rb
+++ b/app/models/concerns/oregon_digital/ability/work_review_ability.rb
@@ -9,7 +9,7 @@ module OregonDigital
       included do
         def work_review_ability
           can :review, SolrDocument do |solr_doc|
-            can_review_solr_doc?(solr_doc.depositor) && can_review_submissions?
+            can_review_solr_doc?(solr_doc.depositor) && can_review_submissions? && is_in_depositors_collection?(solr_doc["edit_access_person_ssim"])
           end
         end
 
--- a/app/models/concerns/oregon_digital/ability/work_review_ability.rb
+++ b/app/models/concerns/oregon_digital/ability/work_review_ability.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal:true
+
+module OregonDigital
+  module Ability
+    # Sets the abilities for work displaying
+    module WorkReviewAbility
+      extend ActiveSupport::Concern
+
+      included do
+        def work_review_ability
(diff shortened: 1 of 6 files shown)


app/models/concerns/oregon_digital/ability/work_show_ability.rb
```diff
@@ -12,7 +12,28 @@ def work_show_ability
...
+            show_record?(record)
+          end
+          can :read, SolrDocument do |solr_doc|
+            read_doc?(solr_doc)
+          end
+        end
+
+        def show_record?(record)
+          if record.suppressed?
+            current_user.role?(admin_permission_roles) && in_depositors_collection?(record.to_solr['edit_access_person_ssim'])
```
CGillen: `record.to_solr['edit_access_person_ssim']` is equivalent to `record.edit_users`

app/models/concerns/oregon_digital/ability/work_show_ability.rb
```diff
@@ -12,7 +12,28 @@ def work_show_ability
...
+          else
+            current_user.role?(manager_permission_roles)
+          end
+        end
+
+        def read_doc?(solr_doc)
+          if solr_doc.suppressed?
+            current_user.role?(admin_permission_roles) && in_depositors_collection?(solr_doc['edit_access_person_ssim'])
```
CGillen: `solr_doc['edit_access_person_ssim']` is equivalent to `solr_doc.edit_people`

app/models/concerns/oregon_digital/ability/work_review_ability.rb
```diff
@@ -0,0 +1,22 @@
...
+module OregonDigital
+  module Ability
+    # Sets the abilities for work displaying
+    module WorkReviewAbility
+      extend ActiveSupport::Concern
+
+      included do
+        def work_review_ability
+          can :review, SolrDocument do |solr_doc|
+            can_review_solr_doc?(solr_doc.depositor) && can_review_submissions? && is_in_depositors_collection?(solr_doc['edit_access_person_ssim'])
```
CGillen: `solr_doc['edit_access_person_ssim']` is equivalent to `solr_doc.edit_people`

PR #789, opened 2019-10-01
[WAIT TO MERGE] updates user and role abilities to match grid

straleyb:
This updates the capabilities of users doing thing in the system around doing User and Role actions. This should be merged later down the road since it will lock down roles and permissions to make it harder to QA. 

WAIT TO MERGE


straleyb:
Merge when ready to finish roles and permissions. Locks down application.

PR #820, opened 2019-10-14
updates view to be a bit more readable

### LINKED ISSUE #694: Frontend work for Roles page
### Descriptive summary

The roles page exists in hyrax but is... less than preferable. We need to work on making the page looking in line with the rest of the application. 

### INDEX

The index page should list the role NOT in a bulleted list. The roles should also not be links, because that is unhelpful. Instead, the Role should be listed, and next to the role should be "Edit" and "Delete" buttons. We may also want this page to 

### SHOW

The show page for a role doubles as the edit page. This doesnt make any sense. We can make a decision whether we want a show page for a role, but on the show page we could display the list of users associated with a role with links to their account show pages. We can also expose an edit and delete button on this page as well, but this will be a small page and we wont be displaying much information besides the name of the role and the users associated. 

### EDIT 

The edit page should contain two sections. One section is where you can edit the name of the role. NOTE: MAYBE WE DONT WANT TO ALLOW THE CHANGING OF THE "ADMIN" ROLE. On that same page we should allow the addition and removal of users from that role. Each user should have a button next to it that says "Remove" and somewhere there should be an autocomplete field which allows a user to search for other users. When one is selected, and "Add" is clicked, it adds the user to the role. 

### MOCKUPS

Currently we dont have any mockups for these pages. Maybe we want some mockups and maybe we dont. @kmthorn @mellinma do we want to create mockups for some of these pages? Or can we take some creative liberty on these as long as it aligns with OD2's current "style"

straleyb:
Fixes #694 

Screen Shot 2019-10-14 at 2 22 11 PM
Screen Shot 2019-10-14 at 2 21 59 PM

### DIFF
--- a/app/views/roles/edit.html.erb
+++ b/app/views/roles/edit.html.erb
@@ -1,0 +1,29 @@
+<h2><%= t('role-management.edit.title') %></h2>
+<%= bootstrap_form_for @role, :url=>role_management.role_path(@role) do |f| %>
+  <%= f.text_field :name, :label=> t('role-management.edit.field_name') %>
+  <%= f.submit t('role-management.edit.update'), class: 'btn btn-primary' %>
+<% end %>
+<h3><%= t('role-management.edit.accounts') %></h3>
+<table class="table table-striped">
+  <thead>
+    <tr>
+      <th> User </th>
+      <th> Action </th>
+    </tr>
+  </thead>
+  <tbody>
+  <% @role.users.each do |user| %>
+    <tr>
+      <td><%= user.user_key %></td>
+      <% if can? :remove_user, Role %>
+        <td><%= button_to t('role-management.edit.remove'), role_management.role_user_path(@role, user.id), :method=>:delete, :class=>'btn btn-danger' %></td>
+      <% end %>
+    </tr>
+  <% end %>
+</table>
+<h3><%= t('role-management.edit.add_new_account') %></h3>
+<%= bootstrap_form_tag :url=> role_management.role_users_path(@role) do |f| %>
+  <%= f.text_field 'user_key', :label=>t('role-management.edit.user') %>
+  <%= f.submit t('role-management.edit.add') %>
+  <%= link_to t('role-management.edit.cancel'), role_management.roles_path, :class => 'btn btn-default' %>
+<% end %>
--- a/app/views/roles/edit.html.erb
+++ b/app/views/roles/edit.html.erb
@@ -1,27 +1,27 @@
-<h2><%= t('role-management.edit.title') %></h2>
+<h1><%= t('role-management.edit.title') %></h1>
 <%= bootstrap_form_for @role, :url=>role_management.role_path(@role) do |f| %>
   <%= f.text_field :name, :label=> t('role-management.edit.field_name') %>
   <%= f.submit t('role-management.edit.update'), class: 'btn btn-primary' %>
 <% end %>
-<h3><%= t('role-management.edit.accounts') %></h3>
+<h2><%= t('role-management.edit.accounts') %></h2>
 <table class="table table-striped">
   <thead>
     <tr>
-      <th> User </th>
-      <th> Action </th>
+      <th scope="col"> User </th>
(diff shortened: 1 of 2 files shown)


straleyb:
This implements the use of tables to make these pages a bit more..... appealing.

app/views/roles/edit.html.erb
```diff
@@ -0,0 +1,29 @@
+<h2><%= t('role-management.edit.title') %></h2>
+<%= bootstrap_form_for @role, :url=>role_management.role_path(@role) do |f| %>
+  <%= f.text_field :name, :label=> t('role-management.edit.field_name') %>
+  <%= f.submit t('role-management.edit.update'), class: 'btn btn-primary' %>
+<% end %>
+<h3><%= t('role-management.edit.accounts') %></h3>
+<table class="table table-striped">
+  <thead>
+    <tr>
+      <th> User </th>
```
CGillen: Needs to be `<th scope="col">` for a11y

app/views/roles/edit.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <th> Action </th>
```
CGillen: Needs to be `<th scope="col">` for a11y

app/views/roles/edit.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    </tr>
+  </thead>
+  <tbody>
+  <% @role.users.each do |user| %>
+    <tr>
+      <td><%= user.user_key %></td>
```
CGillen: Needs to be `<th scope="row">` for a11y

app/views/roles/index.html.erb
```diff
@@ -0,0 +1,21 @@
+<h2><%= t('role-management.index.title') %></h2>
+<table class="table table-striped">
+  <thead>
+    <tr>
+      <th> Role </th>
```
CGillen: Needs to be `<th scope="col">` for a11y

app/views/roles/index.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <th> Action </th>
```
CGillen: Needs to be `<th scope="col">` for a11y

app/views/roles/index.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    </tr>
+  </thead>
+  <tbody>
+    <% @roles.each do |role| %>
+      <tr>
+        <td><%= role.name %></td>
```
CGillen: Needs to be `<th scope="row">` for a11y

app/views/roles/index.html.erb
```diff
@@ -0,0 +1,21 @@
...
+        <td><%= link_to "Manage #{role.name.gsub("_"," ").capitalize} Role", role_management.role_path(role), class: 'btn btn-primary' %></td>
```
CGillen: Text can be simplified to `Manage`. Enough context exists from previous column

app/views/roles/edit.html.erb
```diff
@@ -0,0 +1,29 @@
+<h2><%= t('role-management.edit.title') %></h2>
```
CGillen: Should be `<h1>`, they shouldn't have skipped headings.

app/views/roles/edit.html.erb
```diff
@@ -0,0 +1,29 @@
...
+<%= bootstrap_form_for @role, :url=>role_management.role_path(@role) do |f| %>
+  <%= f.text_field :name, :label=> t('role-management.edit.field_name') %>
+  <%= f.submit t('role-management.edit.update'), class: 'btn btn-primary' %>
+<% end %>
+<h3><%= t('role-management.edit.accounts') %></h3>
```
CGillen: `<h2>`

app/views/roles/index.html.erb
```diff
@@ -0,0 +1,21 @@
+<h2><%= t('role-management.index.title') %></h2>
```
CGillen: `<h1>`

PR #829, opened 2019-10-22
add non-root user app

### LINKED ISSUE #828: Permissions issues when setting up a non-root user
### Descriptive summary:

When deploying the latest build on staging with new USER set to 8083 https://github.com/OregonDigital/OD2/pull/822, we get 
```
`/` is not writable.
Bundler will use `/tmp/bundler/home/unknown' as your home directory temporarily.
```

### Current behavior

To pass this error, bundler ticket https://github.com/bundler/bundler/issues/6663#issuecomment-528179857 suggests setting the HOME env variable to the main app directory.

After applying this change, we get permissions issues such as 

```
Attaching to od2_server_1
server_1           | Building development
server_1           | There was an error while trying to write to `/usr/local/bundle/config`. It is
server_1           | likely that you need to grant write permissions for that path.
```
```
Rails Error: Unable to access log file. Please ensure that /data/log/development.log exists and is writable (ie, make it writable for user and group: chmod 0664 /data/log/development.log). The log level has been raised to WARN and the output directed to STDERR until the problem is fixed.
```

### Expected behavior

It should successfully build containers with new user `8083`.

### Related work
https://github.com/OregonDigital/OD2/pull/822
https://github.com/OregonDigital/OD2/issues/814

### Accessibility Concerns

NA

luisgreg99:
fixes #828

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,28 @@ RUN mkdir -p /opt/fits && \
   curl -fSL -o /opt/fits-1.0.5.zip http://projects.iq.harvard.edu/files/fits/files/fits-1.0.5.zip && \
   cd /opt && unzip fits-1.0.5.zip && chmod +X fits-1.0.5/fits.sh
 
-RUN mkdir /data
+RUN apt-get update && \
+      apt-get -y install sudo
+
+ARG UID=8083
+ARG GID=8083
+
+# Create the home directory for the new app user.
+RUN mkdir -p /data
+RUN chown ${UID}:${GID} /data
+
+# Create an app user so our program doesn't run as root.
+RUN groupadd -g ${GID} app
+RUN useradd -g ${GID} -l -M -s /bin/false -u ${UID} app
+RUN echo "app:app" | chpasswd && adduser app sudo
+RUN echo "app ALL=(root) NOPASSWD:ALL" > /etc/sudoers.d/app && \
+    chmod 0440 /etc/sudoers.d/app
+
+RUN chown -R app:app /data
+
+# Set the home directory to our app user's home.
+ENV HOME=/data
+
 WORKDIR /data
 
 # Pre-install gems so we aren't reinstalling all the gems when literally any
@@ -30,6 +51,7 @@ ADD Gemfile /data
 ADD Gemfile.lock /data
 RUN mkdir /data/build
 
+ENV HOME=/data
 ARG RAILS_ENV=development
 ENV RAILS_ENV=${RAILS_ENV}
 
@@ -37,13 +59,18 @@ ADD ./build/install_gems.sh /data/build
 RUN ./build/install_gems.sh
 
 # Add the application code
-ADD . /data
+COPY --chown=app:app . /data
+
+# Chown all the files to the app user.
+RUN chown -R app:app /data
+
+# Change to the app user.
+USER app
 
 FROM builder
 
 ARG DEPLOYED_VERSION=development
 ENV DEPLOYED_VERSION=${DEPLOYED_VERSION}
-
 
 RUN if [ "${RAILS_ENV}" = "production" ]; then \
   echo "Precompiling assets with $RAILS_ENV environment"; \
@@ -51,8 +78,3 @@ RUN if [ "${RAILS_ENV}" = "production" ]; then \
   cp public/assets/404-*.html public/404.html; \
   cp public/assets/500-*.html public/500.html; \
   fi
-
-ARG UID=8083
-ARG GID=8083
-
-USER $UID:$GID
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -21,7 +21,28 @@ RUN mkdir -p /opt/fits && \
   curl -fSL -o /opt/fits-1.0.5.zip http://projects.iq.harvard.edu/files/fits/files/fits-1.0.5.zip && \
   cd /opt && unzip fits-1.0.5.zip && chmod +X fits-1.0.5/fits.sh
 
-RUN mkdir /data
+RUN apt-get update && \
```
jechols: I think adding sudo is the wrong approach here.  This won't be necessary on production/staging since the image has all the app's code built in.  And while this might be useful during dev, we shouldn't be compromising production to ease development.  (giving the "app" user password-less, restriction-free sudo is only one step away from just running as root)

Dockerfile
```diff
@@ -21,7 +21,28 @@ RUN mkdir -p /opt/fits && \
...
+ARG UID=8083
+ARG GID=8083
+
+# Create the home directory for the new app user.
+RUN mkdir -p /data
+RUN chown ${UID}:${GID} /data
+
+# Create an app user so our program doesn't run as root.
+RUN groupadd -g ${GID} app
+RUN useradd -g ${GID} -l -M -s /bin/false -u ${UID} app
```
jechols: Setting the default command to `/bin/false` seems weird to me; I've not seen this done for user accounts inside images before.  I think we should just stick with existing patterns:

- https://github.com/docker-solr/docker-solr/blob/master/7.7/Dockerfile#L32
- https://github.com/docker-library/mariadb/blob/master/10.3/Dockerfile#L4
- https://github.com/docker-library/redis/blob/master/5.0/Dockerfile#L3

There are also some images which start as root but change their user after doing some setup.  This may be an easier path for us, but I personally haven't looked into that approach.

Dockerfile
```diff
@@ -21,7 +21,28 @@ RUN mkdir -p /opt/fits && \
...
+RUN echo "app:app" | chpasswd && adduser app sudo
+RUN echo "app ALL=(root) NOPASSWD:ALL" > /etc/sudoers.d/app && \
+    chmod 0440 /etc/sudoers.d/app
+
+RUN chown -R app:app /data
+
+# Set the home directory to our app user's home.
+ENV HOME=/data
```
jechols: If we want a home directory, we should just specify that when adding the user (`useradd -d /data` or something).  Or does something not work when HOME is actually assigned via the user?

Dockerfile
```diff
@@ -30,29 +51,30 @@ ADD Gemfile /data
 ADD Gemfile.lock /data
 RUN mkdir /data/build
 
+ENV HOME=/data
```
jechols: This is a duplicate of the above setting

Dockerfile
```diff
@@ -30,29 +51,30 @@ ADD Gemfile /data
...
 
 ADD ./build/install_gems.sh /data/build
 RUN ./build/install_gems.sh
 
 # Add the application code
-ADD . /data
+COPY --chown=app:app . /data
+
+# Chown all the files to the app user.
+RUN chown -R app:app /data
```
jechols: Is this necessary with the above `COPY --chown=app:app` in place?

build/entrypoint.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/sh
 
 echo "Building ${RAILS_ENV}"
+sudo chown -R app:app /data/log
```
jechols: If we need this for development, we should probably introduce a dev entrypoint script.  We definitely shouldn't do this in production since it forces the existence of sudo (see prior note)

docker-compose.yml
```diff
@@ -120,6 +120,7 @@ services:
       - HONEYCOMB_SERVICE=od2
       - HONEYCOMB_WRITEKEY=buzzzzzzzzzzzzzzzz
       - IIIF_SERVER_BASE_URL=http://localhost:8080/iiif
+      - HOME=/data
```
jechols: I don't think this is necessary - it should already be fine in the image, and it shouldn't need to be changed when running the container

build/entrypoint.sh
```diff
@@ -1,6 +1,8 @@
...
+sudo chown -R app:app /usr/local/bundle
```
jechols: This should be in the image definition I think, so that we don't need sudo, and the image is more self-contained.  The production image will already have the gems installed, and if this is a dev-specific thing, we need to find an approach that doesn't compromise production.

jechols:
Might be worth looking into `gosu` if we can't get away from having the container run with escalated privileges - it would be safer than using `sudo`.

jechols:
Closing this - @luisgreg99 is at Samvera Connect, so I usurped his work to solve the core problem (#828)

PR #842, opened 2019-10-28
adds aria lables based on the real label

### LINKED ISSUE #741: Homepage accessibility: search box button label
### Descriptive summary
The dropdown in the search box (pre-filter for 'all works,' 'osu content,' 'uo content,' etc) does not provide a sufficiently descriptive label. Currently, the label is the default value "All." The label should describe the expected input.

### Expected behavior
Searchbox dropdown is labeled as 'collection,' 'scope,' 'filter,' or similar.

### Related work
Audit from #697

### Accessibility Concerns
Labels provide screen readers with hints to announce to the users what the purpose of a form element is. Without a label, the user is left to guess what the element is supposed to do by the context of the possible values (if a mediated field), previous elements, element type, and form function.

straleyb:
Fixes #741

### DIFF
--- a/app/views/catalog/_search_form.html.erb
+++ b/app/views/catalog/_search_form.html.erb
@@ -19,28 +19,34 @@
           <ul class="dropdown-menu">
             <li>
               <%= link_to t("hyrax.search.form.option.all.label_long", application_name: application_name), "#",
-                  data: { "search-option" => main_app.search_catalog_path, "search-label" => t("hyrax.search.form.option.all.label_short") } %>
+                  data: { "search-option" => main_app.search_catalog_path, "search-label" => t("hyrax.search.form.option.all.label_short") },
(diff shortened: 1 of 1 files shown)


straleyb:
Ok, I see. So this is the instruction for whomever using the screen reader to know what that element is for. So "Filter Search Dropdown" They know its a dropdown, they know itll filter their search and they can tab or use the up/down arrow to figure out how they would like to filter their searches.

CGillen:
Yup, that's exactly the idea. I think the aria-labels on the `<li>` elements are redundant at best, but there's an a11y adage of 'if you can do it without aria, do that.' Since the those `<li>`'s have the long_label as the main text in them already I'm pretty sure they'll get recognized by that already.
The only fix should be the aria-label for the dropdown button.

PR #855, opened 2019-10-30
update show page to be in correct order

straleyb:


straleyb:
This is ready for review.

straleyb:
QA: Metadata team can verify that the layout of the show page is in the correct order based on the MAP.

app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -1,26 +1,172 @@
...
-<% end %>
+<% @zipped = {} %> 
 <% Generic.controlled_property_labels.map(&:to_sym).each do |attr| %>
   <% labels=Array(presenter.send(attr)) %>
   <% links=Array(presenter.send(attr.to_s.gsub('_label', ''))) %>
-  <% zipped=labels.zip(links) %>
-  <% zipped=Hash[zipped] %>
-  <%= presenter.attribute_to_html(attr, render_as: :search_and_external_link, links: zipped, html_dl: true, label: t("simple_form.labels.defaults.#{attr}")) %>
+  <% @zipped=labels.zip(links) %>
+  <% @zipped=Hash[@zipped] %>
```
CGillen: This feels like it won't work to present links for controlled vocab metadata below

app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -1,26 +1,172 @@
...
 <% end %>
+
+<%= presenter.attribute_to_html(:trible_title, html_dl: true, label: t("simple_form.labels.defaults.trible_title")) %>
```
CGillen: I wonder if we defined something like `Generic.ordered_properties`, could we repurpose the old loops by just throwing in a check if the property is in the `controlled_property` or `controlled_property_labels`, whichever makes sense.
I don't have a problem with this method, but I think it might be easier to solve the links issue.

straleyb:
Screen Shot 2019-12-03 at 10 35 18 AM

straleyb:
QA: Fill out all fields on the ingest form, including CVs and make sure that all metadata on the show page are in the right order.

PR #875, opened 2019-11-26
Feature/fix solr

jechols:


jechols:
Killing this - need to fix up the branch a tiny bit after noticing a few unrelated commits that should really be discussed separately

PR #878, opened 2019-12-03
Feature/upgrade solr 8

### LINKED ISSUE #818: Upgrade Solr
### Descriptive summary

Solr's alpine image is no longer supported: https://github.com/docker-solr/docker-solr/issues/213

This means we have to update to solr:slim or just the plain solr image.  Maybe we should also bite the bullet and jump to solr 7 at the same time so we have something that's going to be supported a while.

Whatever change we make, it *must* be mirrored in the infra configuration.

### Expected behavior

We have a solid solr setup that is supported.  Rainbows and puppy dogs shoot out of unicorns' faces.  All-around happiness.

jechols:
Updates us to Solr 8.3.  I copied in the default configs from 8.3 and compared them to the prior 8.2 branch, and the only change was that `luceneMatchVersion` field, so I'm pretty confident this is a good move if we were already considering 8.2 a good idea.

Closes #818


jechols:
@luisgreg99 the errors you're describing aren't solr.  The pool timeout is a problem with Fedora (at least that's my current hypothesis, but it's definitely not solr, otherwise we'd have an RSolr timeout).  Label-fetching is completely external to Solr.  I'm not sure if that's the app or Blazegraph, but it's not solr.

straleyb:
@decimalator Any comments on this? You should be aware of these changes just in case any k8s CONFIG_MAPS or anything need to change.

PR #879, opened 2019-12-03
Gem cleanup

### LINKED ISSUE #870: Look at the current rails configuration and trim dev gems
### Descriptive summary

We could look at rails production/development/staging configurations and see if these are the configurations we want to move forward with them. 

We can also look at the dev/staging/production gems and see if we can prune those down too.

CGillen:
Fixes #870


straleyb:
One concern I may have is about fits changes. Do you know if this will have an effect on staging? @decimalator ? 

Other than that, my other concern is about net-http-persistent. I know we dont have an persistent open connections, but im wondering if this had any effect on blazegraph and CVs. Since our triplestore adapter opens connections.

CGillen:
@straleyb Looks like the triplestore adapter already has net-http-persistent ~> 2.9 as a requirement in its gemspec, so we're pulling it in anyways. I was also able to ingest something with a CV field just fine and queried it out of blazegraph afterwards.
I think if we're not going to use it then it probably shouldn't be in our gemfile. I would like to kick `listen` and `coffee-rails` out under the same logic, but their not in the right gemspecs.

I know FITS is used to characterizing files but I don't know what that actually produces in Hyrax so I don't know where to check for changes from 1.0.5. I'm checking what will show up in solr for the fileset of an image on 1.0.5 and 1.5.0 and report back when I can compare

straleyb:
Lets kick them out then, listen and coffee rails, and then push it up and merge it in.

CGillen:
Unfortunately some other gems require coffee-rails and listen, so we'll have to keep them for now.

I'm good with if you want to go forward. Then @jechols can weigh in with more in another PR if he thinks he can slice through them.

PR #884, opened 2019-12-03
Add processor for seed data

jechols:
This was originally a helper for my workflow which I forgot to remove when sharing a branch, and it seems to be helpful for others, so if people want it, here it is.


straleyb:
@luisgreg99 Is this helpful for you? Should we merge this? Curious about what we should do with this PR.

jechols:
We're closing this for now to make sure we don't have a script that gets woefully out of date with the real process.  We'll hold onto the branch in case we decide to reopen this anytime soon.

PR #899, opened 2019-12-09
adds semi-primative generic work with all metadata fields having a value

### LINKED ISSUE #896: Create Rake task to auto-populate work with all metadata fields filled out
### Descriptive summary

This Rake task would quickly create a work that has all metadata fields filled out appropriately (URIs for controlled vocabs, etc.), for testing purposes. Maybe iterate over all of the metadata fields and populate them. 

Can use Example Values from MAP. When new fields are added, either we need a test to make sure this script gets updated too, or we could also have a fallback that would add text to text-only fields. A fallback for controlled fields would be tougher, though as long as our controlled vocabularies and vocabularies don't also change in a commit, that might be doable.

### Expected behavior

Work created with *all* metadata fields filled in, and text or URIs appropriately.

### Related work

This is somewhat related to OD1's filler data task: https://github.com/OregonDigital/oregondigital/blob/master/lib/oregon_digital/filler_data_builder.rb

straleyb:
Fixes #896 

This fills all fields and uses uri's for all controlled fields based on MAP example values.


straleyb:
That was sort of my question. Do we want something right now to just check field order and then refine this to have more proper field values? Or do we want it to be as precise as possible right now? We can talk about it in stand up tomorrow. 

I feel like getting it in now and at least running something just to test that field order PR would be nice. Otherwise this could take another day or two to get in.

wickr:
Another day or two should be fine. 

Although if you can create an object like this with just text in any of the URI fields, that tells me that the controlled vocabularies validation still isn't happening when it needs to be.

straleyb:
This is true. On our backend, there is no validation happening. The app handles the string in a sane way though which is nice. So it indexes the label as a string with no URI. Ill update the script though to detect controlled vocabularies and use the proper uri for it.

lib/tasks/oregon_digital/create_full_metadata_work.rake
```diff
@@ -0,0 +1,187 @@
+# frozen_string_literal: true
+
+namespace :oregon_digital do
+  desc 'Create admin sets in OregonDigital'
```
wickr: This needs updated.

wickr:
The following fields still need to be made to be URIs:
- language
- owner
- publication_place
- workType

I think it's safe to unset values for 'oembed_url' and 'replaces_url' since these are more system level and not descriptive. Otherwise they need to be URLs to save in the ingest form.

Also set Rights Statement and Type, since those are required fields.

With these changes, one can Edit the work and successfully Save.

straleyb:
@wickr I got this updated and editing works. also fills out resource type and rights statement out properly to display.

PR #900, opened 2019-12-10
updated rubocop fixme

straleyb:


lib/hyrax/controlled_vocabularies/location.rb
```diff
@@ -66,7 +66,7 @@ def valid_label_without_parent
       end
 
       def label_or_blank
-        (parentFeature.first.is_a? ActiveTriples::Resource) ? parentFeature.first.rdf_label.first : []
```
CGillen: I remember adding these parens because ruby would not parse this correct for the life of it. It should be checked again if location CVs go into a recursive loop again.

straleyb: Ugh. Ill make a comment on that one that we need it to stick around.

CGillen: I'm not convinced we have to keep it around yet. I'm trying to load up the stack and test it now. That was just my recollection of how that came to be.

It's very possible the ruby upgrade since then fixed it, or I was wrong at the time as far as how necessary it was.

straleyb: Well, fingers crossed. It sucks that we would have to keep something like this around because of that. Lemme know your findings.

CGillen: image
Yup, borked again. BLECH!

straleyb: Ok, Maybe what we do is open a ticket that describes this situation and dig into it? Maybe there are some changes we can make to navigate around this issue. What do you think? Seems like now is a decent time to look at something like this.

CGillen: It might be easier to just rewrite the method quickly here to not use a ternary. I'm not sure what part of it it hates, possibly the double ?'s in the same line or possibly using the open method call syntax (not using parens around parameters, that's probably a new rubocop error).

straleyb: Yeah im not sure either. Its awful suspect though. Smelly. A smelly code smell that smells smelly.

straleyb:
```
return parentFeature.first.rdf_label.first if parentFeature.first.is_a? ActiveTriples::Resource
        
[]
```
Wanna try this?

CGillen: Works and produces the same label. Also acceptable:
```
parentFeature.first.is_a?(ActiveTriples::Resource) ? parentFeature.first.rdf_label.first : []
```

PR #939, opened 2020-01-08
Feature/extract index to job

straleyb:


straleyb:
@jechols Ive taken out the commented out code. It was unnecessary. ill keep those tests xit's out until we know what we want to do with emails. 

Ill get a test up for testing that the code is doing what we want. Im slightly concerned about the SolrService calls, but hopefully I can mock those in some way.

straleyb:
@jechols @CGillen This should be good to go now.

app/indexers/oregon_digital/deep_indexing_service.rb
```diff
@@ -13,13 +13,6 @@ class DeepIndexingService < Hyrax::DeepIndexingService
     private
 
     # OVERRIDEN FROM HYRAX TO ADD GRAPH FETCH RETRY QUEUE
```
CGillen: Update comment to reflect intention of override

app/workers/fetch_graph_worker.rb
```diff
@@ -8,18 +8,51 @@ class FetchGraphWorker
   attr_writer :triplestore
   attr_accessor :uri, :user
 
-  def perform(uri, user)
-    @uri = uri
-    @user = User.find_by_user_key(user)
+  def perform(pid)
+    # @uri = uri
```
CGillen: Remove commented code

spec/models/generic_spec.rb
```diff
@@ -33,11 +33,12 @@
...
     end
 
-    it 'emails the user' do
+    # Note THESE ARE COMMENTED OUT UNTIL WE KNOW WHAT WE WANT WITH EMAILS.
+    xit 'emails the user' do
       expect(Hyrax.config.callback).to receive(:run).with(:ld_fetch_error, user, uri)
       model.send(:enqueue_fetch_failure, uri)
     end
-    it 'enqueues a retry job' do
+    xit 'enqueues a retry job' do
```
CGillen: Remove test or rework for 'enqueues an index job'

spec/workers/fetch_graph_worker_spec.rb
```diff
@@ -5,19 +5,32 @@
   let(:worker) { described_class.new }
   let(:uri) { 'http://my.queryuri.com' }
   let(:user) { build(:user) }
+  let(:model) { create(:generic, title: ['foo'], creator: [controlled_val], depositor: user.email, id: 123) }
+  let(:controlled_val) { OregonDigital::ControlledVocabularies::Creator.new('http://opaquenamespace.org/ns/creator/ChabreWayne') }
 
+  # NOTE: THESE ARE XIT'd out until we know what we want to do with emailing.
   describe '#triplestore' do
-    it { expect(worker.triplestore).to be_a(TriplestoreAdapter::Triplestore) }
+    xit { expect(worker.triplestore).to be_a(TriplestoreAdapter::Triplestore) }
```
CGillen: Shouldn't need to be xit'ed

spec/workers/fetch_graph_worker_spec.rb
```diff
@@ -5,19 +5,32 @@
...
+      stub_request(:get, 'http://opaquenamespace.org/ns/creator/ChabreWayne').with(headers: { 'Accept' => 'application/n-triples, text/plain;q=0.2, application/ld+json, application/x-ld+json, application/rdf+xml, text/turtle, text/rdf+turtle, application/turtle;q=0.2, application/x-turtle;q=0.2, text/html;q=0.5, application/xhtml+xml;q=0.7, image/svg+xml;q=0.4, application/n-quads, text/x-nquads;q=0.2, application/rdf+json, text/n3, text/rdf+n3;q=0.2, application/rdf+n3;q=0.2, application/normalized+n-quads, application/x-normalized+n-quads, text/csv;q=0.4, text/tab-separated-values;q=0.4, application/csvm+json, application/trig, application/x-trig;q=0.2, application/trix, */*;q=0.1', 'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'User-Agent' => 'Ruby RDF.rb/3.0.13' }).to_return(status: 200, body: '', headers: {})
     end
 
-    it 'attempts a local and remote fetch' do
+    xit 'attempts a local and remote fetch' do
```
CGillen: Shouldn't need to be xit'ed, just update `worker.perform()` call

spec/workers/fetch_graph_worker_spec.rb
```diff
@@ -5,19 +5,32 @@
...
       expect(worker.triplestore).to receive(:fetch).twice
       worker.perform(:uri, :user)
     end
+
+    it 'fetches a work and indexes its linked data labels' do
+      work = model
```
CGillen: Rename model to work in `let` definition?

luisgreg99:
I had the chance to test it with a subset of Fairbanks collection (local migration) and with the create_full_metadata_work task, and it worked as expected. When testing, I came across a few fields that didn't fetch the labels (repository, place of production and others), but I think this would be a separate issue since other fields fetched the labels successfully with this new job. I will open a new ticket for this case.

spec/workers/fetch_graph_worker_spec.rb
straleyb: The FetchGraphWorker doesnt actually use a triplestore anymore. So I'll just remove the test.

spec/workers/fetch_graph_worker_spec.rb
straleyb: I dont think this test is necessary anymore either since we are banking on it fetching the values. This is verified via testing that the solr document was updated with labels properly. Since fetch is a butthead.

PR #941, opened 2020-01-09
derivatives: use PNGs for intermediate files

### LINKED ISSUE #937: Fix seed data JP2 derivative failures
### Descriptive summary

Images in the JP2 format are giving us problems in a couple cases, reporting errors like this:

```
gm identify: Improper image header (foo.bmp).
gm identify: Request did not return an image.
```

It's very odd because the BMP files were created successfully by openjpeg, but can't be read by GraphicsMagick.  Odder still is the fact that older versions of GraphicsMagick work with these BMPs, and just produce a warning while still processing and identifying them.

After some investigation, it looks like the tools we have work better when we use TIFF as the intermediate format rather than BMP.  I'll do some testing against a few seed data collections to be sure, but that may end up being the workaround here :-/

### Expected behavior

No valid images give us errors when generating derivatives

jechols:
Fixes our fileset processing service to generate PNGs instead of BMPs
for image derivative intermediate files, which appears to be necessary
for the final conversion from intermediate file to JP2.

It is my believe that this will close #937.  We won't know for certain until we do more testing, but in local dev this worked for all images my system could process (one PDF has 200-megapixel images embedded and my local workstation can't process those)

### DIFF
--- a/app/services/oregon_digital/file_set_derivatives_service.rb
+++ b/app/services/oregon_digital/file_set_derivatives_service.rb
@@ -4,7 +4,7 @@ module OregonDigital
   # Overrides the Hyrax derivative generation for our JP2 setup
   #
   # This builds JP2 derivatives in stages.  The openjpeg tools can't convert
-  # many formats to JP2 directly, so we make sure we have a BMP, and then
+  # many formats to JP2 directly, so we make sure we have a PNG, and then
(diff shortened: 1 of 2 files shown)


luisgreg99:
I tested this PR locally (tried with Baseball_jpegs) and didn't fine any issues.

PR #951, opened 2020-01-14
Feature/doc derivatives

### LINKED ISSUE #938: derivatives process trying to process excel files
although no derivatives are required for this file, df661w36p_content.vnd.ms-excel,  MiniMagick appears to be trying to do something:
MiniMagick::Error: `gm mogrify -flatten /data/mini_magick20200108-1-1915w7j.pdf[0]` failed with error: gm mogrify: Unrecognized option (-flatten).
This is mainly just annoying, because these jobs will need to be cleared out so that sidekiq isn't repeatedly retrying them

jechols:
Closes #938


lsat12357:
tried this locally; the mini-magick error is gone and a thumbnail is now generated.

PR #962, opened 2020-01-27
Set required fields on form to be required on more models

### LINKED ISSUE #950: require identifier in the model
currently, identifier is required in the form, but not the model? This came to light when we migrated seed data assets that don't have an identifier and no errors were thrown. It only became evident when the assets were opened in the front end.

Making identifier required in the model will ensure that regardless of how the asset is generated, identifier will be present.

CGillen:
Fixes #950


straleyb:
@CGillen This needs some rebasing

straleyb:
@CGillen Once this gets rebased, seems like its some simple changes to add them to the rest of the existing models.

PR #965, opened 2020-01-27
Feature/expanded job to use emailing and requeue

straleyb:
This updates the code to fire off a second job if a fetch fails to follow the old way of doing things. Instead of updating everything, it just runs off of a single uri.


straleyb:
I need to test the failed state for the FetchGraphWorker and see if it fires off the new job. Then after that it should be good to go.

app/models/concerns/oregon_digital/work_behavior.rb
```diff
@@ -6,8 +6,11 @@ module WorkBehavior
     extend ActiveSupport::Concern
     include OregonDigital::AccessControls::Visibility
 
+    attr_writer :graph_fetch_failures
```
CGillen: Changes to the work behavior will have no affect if `graph_fetch_failures` doesn't populate. Since you're calling `FetchFailedGraphWorker.perform_async` later, all changes in this file should be safe to drop

app/workers/fetch_graph_worker.rb
```diff
@@ -23,7 +23,12 @@ def perform(pid)
       work.attributes[controlled_prop.to_s].each do |val|
         # Fetch labels
         if val.respond_to?(:fetch)
-          val.fetch(headers: { 'Accept' => default_accept_header })
+          begin
+            val.fetch(headers: { 'Accept' => default_accept_header })
+          rescue TriplestoreAdapter::TriplestoreException
+            fetch_failed_graph(pid, user)
```
CGillen: Should an email go out for individual failures like previously?
`Hyrax.config.callback.run(:ld_fetch_error, user, pid)`

app/workers/fetch_graph_worker.rb
```diff
@@ -45,6 +50,10 @@ def perform(pid)
     ActiveFedora::SolrService.commit
   end
 
+  def fetch_failed_graph(pid, user)
+    FetchFailedGraphWorker.perform_async(pid, user)
```
CGillen: Are pid and uri the same?
I think this might mean what I wrote initially is almost useless, it only dumps the graph into the cache, it doesn't reindex the work.

app/workers/fetch_failed_graph_worker.rb
```diff
@@ -0,0 +1,33 @@
...
+
+# Sidekiq Worker for fetching linked data labels
+class FetchFailedGraphWorker
+  include Sidekiq::Worker
+  sidekiq_options retry: 11 # Around 2.5 days of retries
+
+  attr_writer :triplestore
+  attr_accessor :uri, :user
+
+  def perform(uri, user)
```
CGillen: See previous comment about pid/uri. This perform will probably need to be re-written to re-fetch & re-index the specific field, not just fetch and dump to graph cache. Whoops, sorry.

Issue #963, opened 2020-01-27
task to update protocol on existing location values


luisgreg99:
### Descriptive summary
cleanup to update protocol on existing works with http

### Expected behavior
location on existing assets should include https

### Related work

https://github.com/OregonDigital/OD2/issues/943

### Accessibility Concerns

NA

luisgreg99:
This is ready for QA on staging. Steps for QA:

- [ ] Browse to an existing item with a location: https://staging.oregondigital.org/concern/documents/fx719q28r Location field should be visible
- [ ] Browse to an existing item with location and click "Edit": https://staging.oregondigital.org/concern/documents/fx719q28r/edit?locale=en Under Location, it should list geonames URIs, which should start with `https://`

raybrarian:
QA pass.

Issue #943, opened 2020-01-11
location field not displaying label in the show page


luisgreg99:
### Descriptive summary

Location field doesn't seem to be fetching labels properly in the show pages. The fetching works well in the edit/new forms, but after saving the work, labels don't show up in the view. @wickr found a possible related PR https://github.com/samvera/questioning_authority/pull/307 (it looks like it hasn't been released yet) possibly coming up in an upcoming release of hyrax. The PR suggests that the protocol got changed to https in the response, so https://github.com/OregonDigital/OD2/blob/master/lib/oregon_digital/feature_class_uri_to_label.rb#L6 might need to be updated as well.

### Expected behavior

It should display the location label in the show page.

### Related work
https://github.com/samvera/questioning_authority/pull/307

### Accessibility Concerns

NA

raybrarian:
Hi @luisgreg99 - do you have an example of a work where the Location label should show? I've looked at the "all fields" work visible for UO users (https://staging.oregondigital.org/concern/generics/p5547r392?locale=en) and I do see some location fields, but not one labeled "Location."

luisgreg99:
@raybrarian Yes, here's an example that I've used for testing https://staging.oregondigital.org/concern/images/pz50gw11f?locale=en Also, I had to edit this asset and removed the old location URI, which still included the old `http` protocol (this seems to be the issue in the asset https://staging.oregondigital.org/concern/generics/p5547r392?locale=en as well), in order to fix the display on the show page. 

There is still an issue with existing assets that have the old protocol. For existing assets, I think we can get a script to cleanup these assets. I'll work on this one next.

raybrarian:
@luisgreg99 - ah, got it. It was in review, so I published your test example (https://staging.oregondigital.org/concern/images/pz50gw11f?locale=en#?c=0&m=0&s=0&cv=0&xywh=-272%2C-1%2C1743%2C600) to make sure it looks like it should for public users as part of QA. Looks like the text in Location is showing as well, but is linking to a repository search: 
https://staging.oregondigital.org/catalog?locale=en&q=Eugene+%2C+Lane+%2C+Oregon+%2C+West+%2C+United+States+%2C+Northern+America+%2C+North+America+%2C+Americas+%2C+Earth+++++++%2C+%28%29+&search_field=location_label

So Location shows as a label and has content - do you want me to close this one and the issues we've raised can go in another ticket, or do you want me to boot this one back to backlog?

luisgreg99:
[...]

PR #967, opened 2020-01-28
task to set geonames uris to https

luisgreg99:
fixes https://github.com/OregonDigital/OD2/issues/963
this task covers the case mentioned in https://github.com/OregonDigital/OD2/issues/943#issuecomment-577836324

### DIFF
--- a/lib/tasks/migration/update_location.rake
+++ b/lib/tasks/migration/update_location.rake
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+namespace :migration do
+  desc 'update location'
+  task update_location: :environment do
+    puts 'update location'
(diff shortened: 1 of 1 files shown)


straleyb:
It might be a good idea to extract this code into a service and then test its functionality there.

lsat12357:
Is this more of a remediation task rather than something that needs to go into the repo?

straleyb:
Even if its a remediation task I still think it should go somewhere. If we end up creating 1 new task per remediation, then we could end up with several tasks lying around for unnecessary reasons. Id imagine if the idea is a one off remediation task, then I think it should exist as a script instead of a rake task. But I still think testing it is necessary.

luisgreg99:
It's actually a one time script to fix a specific case (only for seed data) in location field that resulted from a bug already fixed, so I think we can close this one.

PR #968, opened 2020-01-28
make access restrictions controlled

### LINKED ISSUE #959: Make access_restrictions controlled
access_restrictions is controlled according to the OD2 map, but is not set up to be controlled yet in OD2.  waiting on https://github.com/OregonDigital/oregondigital/issues/1306

luisgreg99:
fixes #959

### DIFF
--- a/app/models/concerns/oregon_digital/generic_metadata.rb
+++ b/app/models/concerns/oregon_digital/generic_metadata.rb
@@ -719,7 +719,7 @@ module OregonDigital
         { name: 'rights_note', is_controlled: false },
         { name: 'rights_holder', is_controlled: false },
         { name: 'copyright_claimant', is_controlled: false },
-        { name: 'access_restrictions_label', is_controlled: false },
+        { name: 'access_restrictions_label', is_controlled: true },
         { name: 'identifier', is_controlled: false },
         { name: 'item_locator', is_controlled: false },
         { name: 'accession_number', is_controlled: false },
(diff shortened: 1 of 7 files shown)


app/models/concerns/oregon_digital/generic_metadata.rb
```diff
@@ -719,7 +719,7 @@ module GenericMetadata
         { name: 'rights_note', is_controlled: false },
         { name: 'rights_holder', is_controlled: false },
         { name: 'copyright_claimant', is_controlled: false },
-        { name: 'access_restrictions', is_controlled: false },
+        { name: 'access_restrictions_label', is_controlled: false },
```
straleyb: I think `is_controlled:` should be true here.

straleyb:
Looks good, but could we get some tests for the controlled vocabulary and the endpoint? Should be some examples found here: https://github.com/OregonDigital/OD2/tree/master/spec/oregon_digital/controlled_vocabularies

app/models/concerns/oregon_digital/generic_metadata.rb
luisgreg99: done

luisgreg99:
> Looks good, but could we get some tests for the controlled vocabulary and the endpoint? Should be some examples found here: https://github.com/OregonDigital/OD2/tree/master/spec/oregon_digital/controlled_vocabularies

Ok, I've added specs for the new vocabulary and endpoint

PR #972, opened 2020-01-29
update labels for ordered properties and test

### LINKED ISSUE #960: fields not fetching labels on create and update
### Descriptive summary

I found a few fields that got saved successfully, but the labels didn't get fetched. I found these while testing PR https://github.com/OregonDigital/OD2/pull/939 locally.

- [ ] Form of work
- [ ] Military branch
- [ ] Repository
- [ ] Place of production

### Expected behavior

I think it should fetch the labels and display them in the show page when available.

### Related work

NA

### Accessibility Concerns

None

straleyb:
Fixes #960 

Ill add a screenshot once I get my docker stuff up and working again.


straleyb:
Ok, so I removed the tests. Im having trouble thinking about a way that we could make sure that when a work gets added as a controlled vocabulary, that this list gets updated. One of the issues is that not all controlled vocabularies adhere to the _label piece. In fact there even some displayed cases that have _label that aren't controlled vocabularies. So this list tends to be quite unique.

We could manage a list in the specs as well to check that the ORDERED_PROPERTIES is the same in app as it is in the spec, but that may be completely overkill. Im also not sure how often this list is going to be updated. I say we don't worry about tests right now for this and if this happens another time or two, then we should manage two separate lists in the spec and in the generic_metadata.

Suggestions? @CGillen @luisgreg99 

If you feel creating the list now is worth while, I can definitely do that. I do feel tests are important and will stop us from running into this issue again.

luisgreg99:
@straleyb Another idea would be to sort the current list by the _label suffix and maybe add a comment line describing the difference.

PR #976, opened 2020-01-30
Change required field references to literal list

### LINKED ISSUE #950: require identifier in the model

CGillen:
One
More
Time
Fixes #950


straleyb:
This is good but I think the validates_presence_of needs to be in the included block.

PR #984, opened 2020-01-31
Update fetch_graph_worker.rb

straleyb:


straleyb:
@luisgreg99 This should fix your error. Might actually fix other errors too. Lets get this in.

Issue #981, opened 2020-01-31
FetchGraphJob Randomly sticks when it cant fetch from an endpoint?


straleyb:
### Descriptive summary

Something is happening in FetchGraphJob where it sticks and takes approximately 30 minutes for it to get killed off. It seems to happen either when the endpoint doesn't return fast enough or if there are too many fetches going on. 

Watching the worker logs on staging it seems to blow up on the fetch call and then vomits a huge amount of junk. 

This needs to be dig into deeper to figure out 
1) How many times is Update_index being called?
2) How long are these location fetches taking?
3) How many fetches from geonames is too many in one short period of time?

straleyb:
Sticks because Geonames decides to block the URI. So when geonames removes the limit on us, this shouldnt happen. Closing.

PR #987, opened 2020-02-04
Location indexing

straleyb:
This is more refinement of the FetchGraphWorker. It uses callbacks now instead of firing off on successful save. There is more work needing to be done around this including verification of the indexing.


straleyb:
Screen Shot 2020-02-05 at 7 56 00 AM

straleyb:
Fixes #981

PR #989, opened 2020-02-04
I keep running into issues with this, so temporarily disabling it until we are ready

straleyb:
This will make any wipes of the data and user registration for the time being.


straleyb:
Ill ask if anyone is running into this issue.

PR #993, opened 2020-02-05
Extract string to model so it loads the string when the model is called

### LINKED ISSUE #929: osu key not found bug
### Descriptive summary

After a file is changed, there is some autoload error which doesnt allow for osu key to be found in the form visibility component

straleyb:
Fixes #929 

I updated the code multiple times and the bug didnt show up. I expect this should smooth out development.


straleyb:
@luisgreg99 This one is good to go in too.

Issue #986, opened 2020-02-04
Re-enable emailing in the jobs


straleyb:
We need to inspect why we were getting "Callback not enabled" errors when the jobs were running. So we need to make sure to address the emailing portion as soon as stability for these jobs has increased.

decimalator:
this is dependent on #971 to have something for OD2 to relay email through

PR #1007, opened 2020-02-11
initial refactor of rubocop exceptions

straleyb:


app/workers/fetch_graph_worker.rb
```diff
@@ -5,11 +5,11 @@ class FetchGraphWorker
   include Sidekiq::Worker
   sidekiq_options retry: 11 # Around 2.5 days of retries
 
-  def perform(pid, user_key)
+  def perform(pid, _user_key)
     # Fetch Work and SolrDoc
     work = ActiveFedora::Base.find(pid)
     solr_doc = SolrDocument.find(pid)
-    user = User.where(email: user_key).first
+    # user = User.where(email: user_key).first
```
luisgreg99: I think we can remove the commented out line if it's not needed

app/workers/fetch_graph_worker.rb
```diff
@@ -48,12 +48,12 @@ def perform(pid, user_key)
     ActiveFedora::SolrService.commit
   end
 
-  def fetch_failed_callback(user, val)
-    Hyrax.config.callback.run(:ld_fetch_failure, user, val.rdf_subject.value)
-  end
+  # def fetch_failed_callback(user, val)
```
luisgreg99: We can remove this method as well. Or maybe add a comment to explain the reason for commenting out the method.

spec/workers/fetch_graph_worker_spec.rb
```diff
@@ -29,7 +29,7 @@
       before do
         allow(controlled_val).to receive(:fetch).with(:anything).and_return({})
         allow_any_instance_of(OregonDigital::ControlledVocabularies::Creator).to receive(:solrize).and_return(['http://opaquenamespace.org/ns/creator/ChabreWayne', { label: 'Chabre, Wayne$http://opaquenamespace.org/ns/creator/ChabreWayne' }])
-        allow(worker).to receive(:fetch_failed_callback).with(nil, controlled_val).and_return(true)
+        # allow(worker).to receive(:fetch_failed_callback).with(nil, controlled_val).and_return(true)
```
luisgreg99: Same with this line. I think it can be removed if it's not needed

app/workers/fetch_graph_worker.rb
straleyb: It is. These commented pieces of code will need to be reset. Its on my agenda for #986

app/workers/fetch_graph_worker.rb
straleyb: These commented pieces of code will need to be reset. Its on my agenda for #986

spec/workers/fetch_graph_worker_spec.rb
straleyb: These commented pieces of code will need to be reset. Its on my agenda for #986

app/workers/fetch_graph_worker.rb
luisgreg99: ok sounds good

spec/workers/fetch_graph_worker_spec.rb
luisgreg99: sounds good

app/workers/fetch_graph_worker.rb
luisgreg99: sounds good

PR #1008, opened 2020-02-11
Method length updates

straleyb:


straleyb:
@luisgreg99 @CGillen @jechols So here are some propositions of how we could go about disabling sections of code without adding exclusions in rubocop. Comment why it is disabled and disable only the code that you want to be skipped. There were many pieces of code that couldn't really be shortened, so disabling it seems the best. I think these situations will be far and few between, so I think this is fine. Any other input here?

straleyb:
There are some refactors in here to fix pieces that were needing to actually be fixed.

straleyb:
Also, there were some files that were removed, but still were inside the .fixme file so another positive for doing it in the code rather than in a fixme file.

PR #1056, opened 2020-03-19
Set initial indexing in place for location combined facet

straleyb:
Initial pass at where we need to be indexing Controlled Properties for combined facets. We will be continuing to index them under their initial fields as well as in their new combined facet fields.


straleyb:
Yep

Issue #849, opened 2019-10-29
Search Results: Metadata fields: Add Hit Highlighting


wickr:
### Descriptive summary

Add search keyword hit highlighting to help represent relevance and why search results are returned.

https://github.com/projectblacklight/blacklight/wiki/Configuration---Results-View#highlight

### Expected behavior

For only the following fields: 

* Title is highlighted with any search result hits.
* (IF HIT MATCHES) Description is displayed, in the context of a hit.
* (ONLY IF NO HIT IN DESCRIPTION) Full Text is displayed around the context of a hit.

### Related work

#850 
#851

### Accessibility Concerns

Add any appropriate markup to signify the highlighted keyword in context.

CGillen:
QA:
- [ ] Titles of search results is highlighted when search term is present in title
- [ ] Description is displayed ONLY if there is a hit highlight in it
- [ ] Full Text is displayed and Description is not displayed ONLY if there is a hit highlight in full text (Requires a reindex/resave of a work)

shieldsb:
QA pass.

Issue #850, opened 2019-10-29
Search Results: List View: Metadata fields further update


wickr:
### Descriptive summary

We want relevant and appropriate metadata fields to show up in search results.

### Expected behavior

Show below Title:

* Creator(s)
* Date
* (if keyword hit in Description) Description
* (if NOT keyword hit in Description, AND keyword hit in Full Text) Full Text
* Resource Type (with icon)

(display either Description or Full Text, but not both at same time)

All other fields should not be displayed in List View.

### Related work

#683 

### Accessibility Concerns

Have appropriate markup and attributes to represent the page structure.

CGillen:
Can get most of the way there, but blocked by full text extraction and indexing.
#1055

CGillen:
QA:
@sseymore Please QA or ask someone from metadata to fill in.
Only in the context of the list view
- [ ] Titles of search results is highlighted when search term is present in title
- [ ] Description is displayed ONLY if there is a hit highlight in it
- [ ] Full Text is displayed and Description is not displayed ONLY if there is a hit highlight in full text (Requires a reindex/resave of a work)
- [ ] Metadata displayed in search results includes as expected:
  - [ ] Creator
  - [ ] Date

sseymore:
QA fail (I think!) @CGillen Not sure if this is correct. For "Full Text is displayed and Description is not displayed ONLY if there is a hit highlight in full text (Requires a reindex/resave of a work)" I am seeing description NOT being displayed, but there are also no search results highlighting in Full Text for anything other than OSU test 023.

CGillen:
@sseymore For any work to show up as a search result they need to be resaved or a complete reindex job needs to run. @jechols has a new reindexer in place and we need to run a test on it, but it sounds like we might not do that until the WC is over.

sseymore:
@CGillen Ok, then this is a QA pass- with OSU test 023.

Issue #851, opened 2019-10-29
Search Results: Gallery View: Metadata fields further update


wickr:
### Descriptive summary

We want relevant and appropriate metadata fields to show up in search results.

### Expected behavior

Show below Title:

* Creator(s)
* Date
* (if keyword hit in Description) Description
* (if NOT keyword hit in Description, AND keyword hit in Full Text) Full Text
* Resource Type (with icon)

(display either Description or Full Text, but not both at same time)

All other fields should not be displayed in List View.

### Related work

#687 

### Accessibility Concerns

Have appropriate markup and attributes to represent the page structure.

CGillen:
QA:
@sseymore Please QA or ask someone from metadata to fill in.
Only in the context of the grid view
- [ ] Titles of search results is highlighted when search term is present in title
- [ ] Description is displayed ONLY if there is a hit highlight in it
- [ ] Full Text is displayed and Description is not displayed ONLY if there is a hit highlight in full text (Requires a reindex/resave of a work)
- [ ] Metadata displayed in search results includes as expected:
  - [ ] Creator
  - [ ] Date

sseymore:
See #850 for Full Text is displayed and Description is not displayed ONLY if there is a hit highlight in full text (Requires a reindex/resave of a work)

PR #1057, opened 2020-03-19
Configure search results metadata

CGillen:
This better aligns the search results metadata with #850.
It's not perfect and still needs to display full text. A TODO is added to describe where full text can be put swapped for keyword.


CGillen:
image
Same PDF. First has keyword in description, second does not. PDF text is kind of trash. This isn't about OCR, PR assumes a well formatted PDF

CGillen:
Fixes #849, #850, #851

PR #1074, opened 2020-03-23
Refactor fetch graph worker spec

CGillen:
@straleyb something like this?


straleyb:
Accepted assuming CCI passes

PR #1081, opened 2020-03-24
Feature/update facet order

straleyb:
Screen Shot 2020-03-24 at 2 50 46 PM

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -42,15 +42,10 @@ class CatalogController < ApplicationController
     config.index.display_type_field = solr_name('has_model', :symbol)
     config.index.thumbnail_field = 'thumbnail_path_ss'
 
-    # solr fields that will be treated as facets by the blacklight application
-    #   The ordering of the field names is the order of the display
-    config.add_facet_field solr_name('human_readable_type', :facetable), label: 'Object Type', limit: 5
-    config.add_facet_field solr_name('file_format', :facetable), limit: 5
-    config.add_facet_field solr_name('member_of_collections', :symbol), limit: 5, label: 'Collections'
+    
 
     # The generic_type isn't displayed on the facet list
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -81,7 +74,7 @@ def self.modified_field
       next if Document.properties[prop].behaviors.nil? || Generic.properties.key?(prop)
 
       # Add property as searchable all fields box and individually
-      if Document.properties[prop].behaviors.include?(:stored_searchable)
+      next unless Document.properties[prop].behaviors.include?(:stored_searchable)
```
straleyb: Take note of this change. This used to be an if but it wants a next. Since adding the next, it would skip it if the statement were true. This change SHOULD fix that. More testing is needed though

straleyb: Strike that. I need to dig into this further. But something has fundamentally changed about this statement that I need to discover.

app/controllers/catalog_controller.rb
```diff
@@ -42,15 +42,8 @@ def self.modified_field
...
 
-    # solr fields that will be treated as facets by the blacklight application
-    #   The ordering of the field names is the order of the display
-    config.add_facet_field solr_name('human_readable_type', :facetable), label: 'Object Type', limit: 5
-    config.add_facet_field solr_name('file_format', :facetable), limit: 5
-    config.add_facet_field solr_name('member_of_collections', :symbol), limit: 5, label: 'Collections'
-
     # The generic_type isn't displayed on the facet list
     # It's used to give a label to the filter that comes from the user profile
-    config.add_facet_field solr_name('generic_type', :facetable), if: false
```
CGillen: Judging by the comment, I think this is used somewhere in admin pages. The `if: false` keeps it from being displayed but I think blacklight needs to know about `generic_type` during its facet query to return some data for somewhere else in Hyrax

PR #1087, opened 2020-03-26
update RESOURCE_TYPE not TYPE

straleyb:


CGillen:
Before you go mucking with tests you might want to come at this fix from a different angle

https://github.com/OregonDigital/OD2/blob/master/app/controllers/catalog_controller.rb#L163
`I18n.translate('simple_form.labels.defaults.resource_type')` 
the translation can be updated to the right one `I18n.translate('simple_form.labels.defaults.type')`

PR #1111, opened 2020-04-07
Display IIIF viewer if child has IIIF viewable images

### LINKED ISSUE #998: A parent work should show universal viewer for child works it has even if the parent doesn't have a fileset associated
### Descriptive summary

Universal viewer depends on a work having a file set in order to show universal viewer. When works are migrated, the parent works don't have file sets associated with it. We still want to show universal viewer for a parent work even if it does not have a file set to display its children in universal viewer.

CGillen:
Fixes #998


CGillen:
A parent work with 2 children and no filesets:
test

Parent with 1 child and existing fileset
Screen Shot 2020-04-07 at 2 08 57 PM

PR #1125, opened 2020-04-14
Stop geonames parent search when rdf_subject is https

### LINKED ISSUE #1124: Update Location URI Labels and hierarchy
### Descriptive summary

Labels for Location URIs referencing GeoNames should provide enough information in the hierarchy to disambiguate, but not too much that the strings are too long. GeoNames provides the hierarchy structure, but we need to place some limits to ensure that only the information we want is included.

We want the hierarchy display to stop at Country. This ends up being 3 (added) levels max. (it looks like this should be working now, but maybe the GeoNames response update to include 'https' is breaking the `top_level_codes` check)

Continents and other higher-level values should not display as part of the hierarchy, but are fine if that is the URI put in: Asia, North America, Earth, etc.

When going up the hierarchy, make sure to *only* use the 'Administrative hierarchy' values, not the 'Parents' values. For example, for Oregon: https://www.geonames.org/5744337/oregon.html
(click 'Hierarchy' button) see the 'Administrative hierarchy' has United States listed as country. But click over to 'Parents' and see 'West', 'Pacific Northwest' and 'Pacific states', which are all ones we do not want to display or follow.

Feature code labels should not display.

County names should include 'County': Deschutes County

### Expected behavior
(several examples here, not meant to be exclusive)

valid forms:
  - [City] >> [County] >> [State] >> United States
  - [River] >> [County] >> [State] >> United States
  - [County] >> [State] >> United States
  - [State] >> United States
  - [City] >> [Province] >> [Country]
  - [Country]

also valid:
  - Northwest >> United States
  - Pacific Northwest >> United States
  - Asia

### Accessibility Concerns

N/A

CGillen:
Fixes #1124 
image

### DIFF
--- a/lib/hyrax/controlled_vocabularies/location.rb
+++ b/lib/hyrax/controlled_vocabularies/location.rb
@@ -69,7 +69,7 @@ module Hyrax
         parent_country = parentCountry.first
         sec_adm_level_code = 'www.geonames.org/ontology#A.ADM2'
         us_country_code = 'sws.geonames.org/6252001/'
-        feature_code.responds_to?(:rdf_subject) && feature_code.rdf_subject.to_s.include?(sec_adm_level_code) && parent_country.responds_to?(:rdf_subject) && parent_country.rdf_subject.to_s.include?(us_country_code)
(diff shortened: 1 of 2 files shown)


straleyb:
@CGillen ^^^ Was this addressed?

CGillen:
Nope, continuing work

CGillen:
@wickr 
Ignore the double metadata. I reindexed twice too quickly
image

CGillen:
Third time
image

PR #1127, opened 2020-04-14
update to make confirmation decision

straleyb:
Refactor + confirmation notification check.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -58,7 +58,6 @@ class User < ApplicationRecord
       u.email = email
       u.roles << role unless role.nil?
       u.skip_confirmation!
-      u.skip_confirmation_notification!
     end
   end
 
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -37,6 +37,13 @@ def mailboxer_email(_obj = nil)
     email
   end
 
+  # Override from devise to make a determination of whether to send email notification or not
+  def send_confirmation_notification?
```
CGillen, line 41: Looks like `#first_or_create` from below calls save, so calling `#skip_confirmation!` inside the create block should be enough to cover everything. I don't think we need the override for `#send_confirmation_notification?`

straleyb: We do. Margaret tested it and apparently skip_confirmation_notification! wasn't actually being called at the right time. So instead of skip_confirmation_notification! we can use #send_confirmation_notification? override to get it to make a determination.

CGillen: Alright, do you want to simplify it down to a one-liner and maybe include `#super` so it's still going through devise's checks? Something like 
`super && !(email.include?('uoregon') || email.include?('oregonstate'))`
or 
`super && !email.include?('uoregon') && !email.include?('oregonstate')`

PR #1130, opened 2020-04-16
bring in input and call fetch so rdf_label works

straleyb:


straleyb:
@CGillen so here is the question. Do we override this file and call fetch, or do we create our own input and switch all our inputs to the new one which calls fetch?

straleyb:
Screen Shot 2020-04-16 at 10 37 45 AM

straleyb:
https://github.com/OregonDigital/OD2/pull/1130/files#diff-0e2e612a8d642cf8ffa10734d378d005R69 Line for reference

CGillen:
We already have a new class for single value controlled vocab inputs. It's probably better to make our own, update the single value inheritance and change all our inputs to our new one

straleyb:
Are you suggesting making our own `controlled vocabulary input` and then have the `single value controlled input` inherit from our own `controlled vocabulary input`? I wouldn't want to use the `single value controlled input` across the board since its meant for a few specific fields.

CGillen:
Correct and same. I just wanted to mention updating the single value controlled input so it doesn't get missed.
It would be nice if we somehow didn't have to fetch all these values, but I'm not sure how that would work. It's a MAJOR slowdown for the form

straleyb:
So thats what I was playing with. 

Here is what my list of ideas were
- [ ] Grabbing the solr document and pulling the label from the solr document. 
      (This would hit Fedora and we would need some way to truck around the id to grab it)
- [ ] Fetching using the triplestore instead of the value itself. 
      (This was convoluted since we would have to do parsing of the graph and stuff)
- [ ] Update the value in the build to the value of the _label
      (This is still hitting fedora for the SolrDocument and also overriding other pieces that call #build_field)

This at least is for sure to hit the triplestore since the values have already been fetched since these fields are only in this mode if a field exist.

straleyb:
I think the struggle is the labels are only indexed and this is pulling from the object itself. (Object via the form object f.object is an actual Work Object)

CGillen:
New screenshots
image

PR #1140, opened 2020-04-25
update hyrax-migrator 6-3-2020

luisgreg99:
gem updates:

- https://github.com/OregonDigital/hyrax-migrator/pull/197 by lsat12357  :
  - precheck tool based on crosswalk and admin services 
  - removed function `attributes_replaces_data` for predicate http://purl.org/dc/terms/rights 
- [https://github.com/OregonDigital/hyrax-migrator/pull/206](https://github.com/OregonDigital/hyrax-migrator/pull/206) by dependabot bot 
  - Bump puma from 4.3.3 to 4.3.5  dependencies
- [https://github.com/OregonDigital/hyrax-migrator/pull/203](https://github.com/OregonDigital/hyrax-migrator/pull/203) by luisgreg99
  - verify derivatives service


luisgreg99:
> I think we still need to catch the dct:rights predicate in the overrides, just not use the function?

@lsat12357 Good catch, I restored the predicate in the overrides and removed the function only.

luisgreg99:
With the latest hyrax-migrator update, rdf was bumped to from 3.1.1 to 3.1.2. I'm also updating resource_spec to address the spec error with rdf 3.1.2:
```
ArgumentError: expected graph_name to be nil or a resource, was "bg"
```
It looks like the graph_name argument should now be nil or a valid URI to be valid for this new version. Rdf 3.1.2 throws this ArgumentError for invalid graph name: https://github.com/ruby-rdf/rdf/blob/1a33e000c1cef9a3c588ba56dec1957582cc467a/lib/rdf/model/statement.rb#L134

PR #1160, opened 2020-05-27
Allow user to create user collection

### LINKED ISSUE #1143: [MyShelf] Allow an unprivileged authenticated user to create collections
### Descriptive summary
As part of My Shelf functionality, a user should be allowed to create a collection (the shelf) of a `User Collection` type.
User collections should require no more metadata input than title and description.

### Expected behavior
Any authenticated user can create a personal collection.

### Related work
Subticket of:
#643

straleyb:
Fixes #1143

### DIFF
--- a/app/models/concerns/oregon_digital/ability/collection_ability.rb
+++ b/app/models/concerns/oregon_digital/ability/collection_ability.rb
@@ -9,8 +9,7 @@ module OregonDigital
       included do
         def collection_ability
           can(%i[index show new create edit update delete], Hyrax::CollectionType) do collection_type
-            true if collection_type.title == 'User Collection'
-            current_user.admin?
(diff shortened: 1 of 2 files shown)


app/models/concerns/oregon_digital/ability/collection_ability.rb
```diff
@@ -8,9 +8,18 @@ module CollectionAbility
 
       included do
         def collection_ability
-          can(%i[index show new create edit update delete], Hyrax::CollectionType) if current_user.admin?
+          can(%i[index show new create edit update delete], Hyrax::CollectionType) do collection_type
+            if collection_type.title == "User Collection"
```
CGillen: Two things. I think `collection_type.machine_id == :user_collection` is more robust
and this should simplify to `collection_type.machine_id == :user_collection || current_user.admin?`

PR #1161, opened 2020-05-28
override sortandperpage and add modal

straleyb:
Initial work for modal on search results page.

Still need to fix some CSS and make the button text toggle when one search result is present vs many


straleyb:
Screen Shot 2020-05-29 at 8 26 27 AM
Screen Shot 2020-05-29 at 8 27 22 AM

PR #1197, opened 2020-06-16
Update to add callback to updating a member

straleyb:
Here is an initial proposal to fixing the metadata disappearing after a work is added to a collection. One thing to note here is there is a reindex that happens when a work is added to a collection. this means that metadata (specifically controlled vocabularies) may not be available instantly after a work is added to a collection. 

This has been noted previously when a work is created. But this needs to be extended to "anytime a work is saved" which may not just be on create/update. 

Therefore we have to be mindful of the state of an application after someone adds a work to a collection.

### DIFF
--- a/config/initializers/hacks/hyrax_collections_members_controller.rb
+++ b/config/initializers/hacks/hyrax_collections_members_controller.rb
@@ -15,4 +15,23 @@ Hyrax::Dashboard::CollectionMembersController.class_eval do
       format.json { render json: @collection, status: :updated, location: dashboard_collection_path(@collection) }
     end
   end
+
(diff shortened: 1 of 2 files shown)


config/initializers/hacks/hyrax_collections_members_controller.rb
```diff
@@ -15,4 +15,23 @@ def after_update
       format.json { render json: @collection, status: :updated, location: dashboard_collection_path(@collection) }
     end
   end
+
+  def update_members
```
straleyb, line 19: Note, this was the proper place to add the callback. Other place wasnt correct.

PR #1206, opened 2020-06-19
lower duration of time for commits and commit sooner

### LINKED ISSUE #1205: Update the fetch job to commit earlier in the mechanism.
### Descriptive summary

Instead of making one large commit, we can make incremental commits to make the data accessible quicker.

straleyb:
Fixes #1205

### DIFF
--- a/app/workers/fetch_graph_worker.rb
+++ b/app/workers/fetch_graph_worker.rb
@@ -23,7 +23,6 @@ class FetchGraphWorker
     solr_doc['_version_'] = 0
 
     # Iterate over Controller Props values
-    # rubocop:disable Metrics/BlockLength
     work.controlled_properties.each do |controlled_prop|
       work.attributes[controlled_prop.to_s].each do |val|
         begin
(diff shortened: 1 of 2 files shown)


app/workers/fetch_graph_worker.rb
```diff
@@ -47,22 +47,22 @@ def perform(pid, _user_key)
...
           else
             extracted_val = val.solrize.last.is_a?(String) ? val.solrize.last : val.solrize.last[:label].split('$').first
             Solrizer.insert_field(solr_doc, "#{controlled_prop}_label", [extracted_val], behavior)
             Solrizer.insert_field(solr_doc, 'location_combined_label', [extracted_val], behavior) if location_combined_facet?(controlled_prop)
             Solrizer.insert_field(solr_doc, 'creator_combined_label', [extracted_val], behavior) if creator_combined_facet?(controlled_prop)
             Solrizer.insert_field(solr_doc, 'topic_combined_label', [extracted_val], behavior) if topic_combined_facet?(controlled_prop)
             Solrizer.insert_field(solr_doc, 'scientific_combined_label', [extracted_val], behavior) if scientific_combined_facet?(controlled_prop)
+            ActiveFedora::SolrService.add(solr_doc)
+            ActiveFedora::SolrService.commit
           end
```
CGillen, line 61: I think maybe after this line (the if,else,end block) would be a better place for those.

Also, we might want to consider simplifying the branching
```
val = (val.solrize.last.is_a?(String) ? [val.solrize.last] : [val.solrize.last[:label].split('$').first]) unless val.is_a?(String)
```

PR #1219, opened 2020-06-25
Filters user collections from facet item list

### LINKED ISSUE #1174: Remove User Collections from search results
### Descriptive summary
A user can see their private collections, and other user's public collections from the search results. We need to keep the search results clean of user collections because they have no curated grouping and may clutter the user experience.

### Expected behavior
User Collection collection types do not return in public search results under any conditions.

straleyb:
Fixes #1174

### DIFF
--- a/app/helpers/oregon_digital/facet_helper_behavior.rb
+++ b/app/helpers/oregon_digital/facet_helper_behavior.rb
@@ -1,0 +1,13 @@
+module OregonDigital::FacetHelperBehavior
+  def render_facet_item(facet_field, item)
+    if facet_in_params?(facet_field, item.value)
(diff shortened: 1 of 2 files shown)


app/helpers/oregon_digital/facet_helper_behavior.rb
```diff
@@ -0,0 +1,13 @@
+module OregonDigital::FacetHelperBehavior
+  def render_facet_item(facet_field, item)
```
CGillen, line 2: Can you add an `OVERRIDE FROM BLACKLIGHT` comment here? So we know what to look back on if this method changes

straleyb: 👍

PR #1268, opened 2020-08-10
fixes to crosswalk

lsat12357:
fixes https://github.com/OregonDigital/hyrax-migrator/issues/224


config/initializers/migrator/crosswalk_overrides.yml
```diff
@@ -1,58 +1,76 @@
...
+    predicate: http://opaquenamespace.org/rights/rightsHolder
     multiple: true
+
+  #multiple must be true? (is false on map)
   - property: repository_attributes
     predicate: http://id.loc.gov/vocabulary/relators/rps
     multiple: true
     function: attributes_data
+
+  #duplicates the main entry?
```
luisgreg99, line 31: I think this is a good catch! I think we should remove `creator_attributes`, `local_collection_name_attributes`, and `subject_attributes` from the overrides. I double checked with what we have in croswalk.yml, and I confirm they are duplicates. 

Now that I double checked, these got added in https://github.com/OregonDigital/OD2/pull/703 to the overrides to support the nested attributes syntax in OD2. It looks like https://github.com/OregonDigital/OD2/pull/703 got merged before our decision to reset crosswalk and sync it with the MAP, including setting the nested attributes syntax in crosswalk.yml instead of doing it in crosswalk.overrides.yml. This change happened in https://github.com/OregonDigital/OD2/pull/788, but I think we missed to remove them in the overrides.

PR #1287, opened 2020-08-24
If no university login, then redirect to homepage

### LINKED ISSUE #1232: Public user login should redirect to homepage
### Descriptive summary
If a user has no roles or is only institutionally affiliated, they should be brought to the homepage, not the dashboard

### Expected behavior
non-admin users never see the admin dashboard

straleyb:
Fixes #1232

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -5,12 +5,19 @@
 class Users::SessionsController < Devise::SessionsController
   before_action :redirect_if_university, only: [:create]
 
+  def create
+    self.resource = warden.authenticate!(auth_options)
+    set_flash_message!(:notice, :signed_in)
+    sign_in(resource_name, resource)
+    yield resource if block_given?
+    respond_with resource, location: '/'
+  end
+
   protected
 
   def redirect_if_university
     service = OregonDigital::UserAttributeService.new(params[:user])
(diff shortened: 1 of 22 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -11,5 +11,6 @@ def redirect_if_university
     service = OregonDigital::UserAttributeService.new(params[:user])
     redirect_path = service.email_redirect_path
     redirect_to redirect_path unless redirect_path.nil?
```
CGillen, line 13: Does this still redirect institutional users without admin, depositor, or collection manager roles to the dashboard?

straleyb: oh good question. ill check

straleyb: Answer: Yes, logging in via institutional login without admin depositor or collection manager still redirects to dashboard

CGillen: I think we want only those three roles to hit the dashboard, everyone else should go to the homepage or the page they left from (if easy)

straleyb: ahhh ok i think the title to the ticket might have misled me. Ill swing back around and verify that everyone except admin/collectionmanagers/depositors route to homepage. Else go to dashboard.

PR #1292, opened 2020-08-27
Collections landing page wireframe implementation

### LINKED ISSUE #43: [10] Design for Collections Landing Pages
### Descriptive summary

This may or may not be work done in hyrax. There should be the ability to express a collection landing page per collection. This landing page would contain information about the collection. Whats in it? Where did it come from? What is its significance? Some sort of description of the collection. It should also have representative images associated with the collection. ~These should be displayed on the landing page in a carousel like rotation.~ These should be displayed in a single row of thumbnails, 4 across.

These have already been implemented in OD and are a very important piece of displaying a collection for users to see. This collection landing page needs to be accessible from a list of collections somewhere in OD2. 

### Expected behavior

When the collection is visited, all the information associated with the collection will be displayed on its landing page.

Latest MAC design:
https://projects.invisionapp.com/share/47X400SV3BM#/screens/415938473

###Related Issues
#42

CGillen:
Fixes #43 

This is where we're at:
image

Desktop wf: https://projects.invisionapp.com/share/47X400SV3BM#/screens/415938474
Mobile wf: https://projects.invisionapp.com/share/YJX9Y48C9H3#/screens/417551193

My next moves was to work on the representative images masonry. My idea was to pull from the collections controller method `#files` to get the first 4 images associated with the collection and add them to a new views structure. This could be improved by grabbing the file from `#thumbnail_id` first, as that will be the (or many if we fix it later for user defined representative images) representative image.

My idea for views structure is something like:
- `_masonry_showcase.html.erb` - Contains masonry structure, may require some JS work to get gutters and items per line right
  - `_showcase_tile.html.erb` - Contains the image and hover text. Use CSS to hide text and `:hover` to display.
See the Explore Collections setup to pull some examples and ideas how this could work.

Also, I've been using that `$brand-primary: $apple-green;` line in `_variables.scss` as a cache buster, it seems like the sass cache doesn't really bust for anything but the top level files, and you may need to delete `tmp/cache` every so often. Just remember to remove the apple-green line before we merge.


straleyb:
Screen Shot 2020-09-01 at 11 17 38 AM

app/assets/stylesheets/oregon_digital/explore_collections/_showcase.scss
```diff
@@ -2,4 +2,14 @@
   padding-top: 2em;
   border-top: 3px solid $very-light-grey;
   background-color: unset;
+}
+
+.apple-green-hr {
+  border: 2px solid #D0DD28;
```
CGillen, line 8: Please use `$apple-green` for the color

app/views/hyrax/collections/_showcase_tile.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="thumbnail">
+    <div id="image-<%= index %>" onmouseover='toggleTitleOn("<%= index %>")' onmouseout='toggleTitleOff("<%= index %>")'>
```
CGillen, line 2: Here's some thoughts on hover functionality:
We CAN do it with JS and that's probably easier conceptually and maybe for maintenance. 
But we could also use the `:hover` CSS selector to change CSS, rather than add/remove style/classes with mouse events. This will work more consistently across browsers and may perform better.

Something like this should work:
```
img.hoverable {
  border: none;
  &:hover { // & Means the parent selector. Simplifies to "img.hoverable:hover"
    border: 5px solid $black;
  }

  & + .show-me { // + Means sibling to. Simplifies to "img.hoverable (find adjacent) .show-me"
    margin-top: -1em; // Just assuming the margins can be tweaked down to em relative, not pixel perfect. Not a big deal if not, may cause mobile problems if using pixels?
    margin-left: -2em;
    color: white;
    display: none;
  }
  &:hover + .show-me {
    display: block;
  }
}
```

I don't know at this point if where you'd want to put that CSS and if we want to make it more specific or if it's going to be used again for similar cards. I think it could be used again on the search results pages.

app/views/hyrax/collections/show.html.erb
```diff
@@ -1,73 +1,89 @@
...
-            <% unless @presenter.modified_date.blank? %>
-                <div class="hyc-last-updated">Last Updated: <%= @presenter.modified_date %></div>
-            <% end %>
-          </div>
+<div class="hyc-container" itemscope itemtype="http://schema.org/CollectionPage">
+  <div class="row">
+    <div class="hyc-blacklight hyc-bl-results">
+      <div class="row">
+      <div class="col-md-2"></div>
+      <% @member_docs[0,4].each_with_index do |doc, i| %>
```
CGillen, line 42: Definitely doesn't need to happen here, but maybe we can rework `@member_docs` to take into account the `thumbnail` metadata field. It's looking like `thumbnail` can be used for the collection "feature works" functionality, and if we can make `@member_docs` aware of that, we can make the first N works always be from `thumbnail`. That would give us a pretty consistent way to access featured works or first N works.

app/views/hyrax/collections/show.html.erb
```diff
@@ -1,73 +1,89 @@
...
+  <!-- Search bar -->
+  <div class="hyc-blacklight hyc-bl-search hyc-body row">
+    <div class="col-sm-4"></div>
+    <div class="col-sm-4">
+      <%= render 'search_form', presenter: @presenter, url: hyrax.collection_path(@presenter.id) %>
     </div>
+    <div class="col-sm-4"></div>
+  </div>
+
+  <div class='row'>
```
CGillen, line 76: We use these rows a few places in the wire frames. Consider them for a refactor of template files

app/views/hyrax/collections/_showcase_tile.html.erb
straleyb: Resolved. this is ok but we can look back and see what pieces of the app could benefit from being JS-less down the road to help further with accessibility.

CGillen:
Approved, @straleyb will have to give the actually GH approval

Issue #1365, opened 2020-11-05
Replace calls to DataTables.js with custom pagination solution


CGillen:
### Descriptive summary
Small pagination feels better to the user if done in JS on the front end. DataTables.js provided enough pagination functionality to get off the ground, but required a `<thead>` and `<th>` elements to function. Those are poor for accessibility when the table is being used as a layout table.

We should replace calls to DataTables.js with some custom JS that follows the same idea to build a _layout_ table with pagination

### Expected behavior
Pagination still works for compound object show pages, but datatables.js isn't used, or more importantly the `thead` and `th` aren't in the layout table.

### Accessibility Concerns
This is a ticket about improving a11y

CGillen:
QA:
- [ ] Visit a parent object: https://staging.oregondigital.org/concern/generics/df727084j?locale=en
- [ ] Page loads
- [ ] 4 child works appear
- [ ] Paginator scrolls through works
- [ ] Visit one of the child works
- [ ] 1 parent work (above) appears
- [ ] 4 sibling works appear
- [ ] Sibling paginator scrolls through sibling works, but not parent work

PR #1462, opened 2021-01-13
updates to paginate using datatables

straleyb:


CGillen (approved):
Another one for #1365 to clean up

PR #1505, opened 2021-02-01
lock added with tooltip. Needs refinement

straleyb:


app/assets/stylesheets/oregon_digital/explore_collections/_buttons.scss
```diff
@@ -68,4 +68,12 @@
   position: relative;
   margin-left: 55px;
   visibility: hidden;
+}
+
+.hide {
```
CGillen, line 73: Use `.hidden` instead of adding a new class

app/views/oregon_digital/explore_collections/_list_showcase.html.erb
```diff
@@ -13,8 +13,19 @@
   </thead>
   <tbody>
     <% collections.each_with_index do |collection, i| %>
+      <% institution = collection.institution.first == "http://id.loc.gov/authorities/names/n80017721" ? "OSU" : "UO" %>
+      <% institution = "" if collection.institution.empty? %>
+      <% if Collection.find(collection.id).insitution_restricted?(current_ability) %>
```
CGillen, line 18: Maybe save the restricted status so we don't have to do 2x Fedora lookups (even if they're cached)

CGillen (changes requested):

straleyb:
Screen Shot 2021-02-02 at 12 55 47 PM

CGillen (approved):

PR #1508, opened 2021-02-03
get facet reporter, tmpdir fix

lsat12357:
get updated migrator (contains https://github.com/OregonDigital/hyrax-migrator/issues/255, https://github.com/OregonDigital/hyrax-migrator/issues/240)


lsat12357:
I think I figured out how to just get the migrator update without updating everything else

CGillen (approved):

PR #1511, opened 2021-02-03
Feature/refactor institution checks

straleyb:


app/services/oregon_digital/institution_picker.rb
```diff
@@ -0,0 +1,49 @@
...
+        institution == osu ? 'OSU' : 'UO'
+      when :full
+        institution == osu ? 'Oregon State University' : 'University of Oregon'
+      else
+        ''
+      end
+    end
+
+    def self.contains_university?(collection)
+      collection.institution.include?(uo) || collection.institution.include?(osu)
```
CGillen, line 38: Maybe we can do this for easier extension later?
`collection.institution.intersection(uo + osu).empty?`

My only concern is if there ends up being more URIs that get used to point to OSU/UO. I've seen one for dbpedia or something similar.

CGillen: This obviously makes a lot of other logic in here need some changes to revolve around arrays

app/services/oregon_digital/institution_picker.rb
```diff
@@ -0,0 +1,49 @@
...
+      return '' unless contains_university?(collection)
+
+      institution = selected_institution(collection)
+      translate_institution(institution, :full)
+    end
+
+    def self.selected_institution(collection)
+      institution = collection.institution.select { |val| val == osu }.first
+      institution = collection.institution.select { |val| val == uo }.first if institution.blank?
+      institution
```
CGillen, line 23: `collection.institution.intersect(osu + uo).first`

app/services/oregon_digital/institution_picker.rb
```diff
@@ -0,0 +1,49 @@
...
+        ''
+      end
+    end
+
+    def self.contains_university?(collection)
+      collection.institution.include?(uo) || collection.institution.include?(osu)
+    end
+
+    def self.osu
+      'http://id.loc.gov/authorities/names/n80017721'
```
CGillen, line 42: Then just make these arrays and we can add new URIs later

CGillen (changes requested):

app/services/oregon_digital/institution_picker.rb
```diff
@@ -0,0 +1,49 @@
...
+    def self.selected_institution(collection)
+      institution = collection.institution.select { |val| val == osu }.first
+      institution = collection.institution.select { |val| val == uo }.first if institution.blank?
+      institution
+    end
+
+    def self.translate_institution(institution, translation_type)
+      case translation_type
+      when :acronym
+        institution == osu ? 'OSU' : 'UO'
```
CGillen, line 29: `osu.include?(institution)`

app/services/oregon_digital/institution_picker.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    def self.selected_institution(collection)
+      institution = collection.institution.select { |val| val == osu }.first
+      institution = collection.institution.select { |val| val == uo }.first if institution.blank?
+      institution
+    end
+
+    def self.translate_institution(institution, translation_type)
+      case translation_type
```
CGillen, line 27: Maybe a guard clause to check that `institution` is in either `#uo` or `#osu`

CGillen (approved):

PR #1520, opened 2021-02-08
updates to filter institutional lock based on works

straleyb:


app/controllers/oregon_digital/explore_collections_controller.rb
```diff
@@ -46,9 +46,29 @@ def my_collections
...
+    def osu_items(id)
+      Hyrax::SolrService.query("member_of_collection_ids_ssim:#{id} AND visibility_ssi:osu")
+    end
+
+    def uo_items(id)
+      Hyrax::SolrService.query("member_of_collection_ids_ssim:#{id} AND visibility_ssi:uo")
+    end
+
+    def osu_restricted?(id)
+      !osu_items(id).empty? && current_ability.cannot?(:show, osu_items(id).first, visibility: 'osu')
```
CGillen, line 61: As a note for later, we might run into a situation were the first item is not representative of all OSU/UO works. That's why I was hoping the `current_ability.cannot?(:show, ActiveFedora::Base, visibility: 'osu')` syntax would work.

We might have to do this to test all works, which may be slow: `current_ability.cannot?(:show,  {:any => osu_items(id)})`
Alternatively: 
```
dummy = ActiveFedora::Base.new
dummy.visibility = 'osu'
current_ability.cannot?(:show, dummy)
```

CGillen (approved):

PR #1607, opened 2021-04-09
Migrator config files updates

lsat12357:
fixes https://github.com/OregonDigital/hyrax-migrator/issues/270
additionally brings map updates to the two crosswalk files


lsat12357:
I just ran preflight using the crosswalk files, so far so good

wickr:
closed in favor of #1616

PR #1616, opened 2021-04-14
Migrator config updates

### LINKED ISSUE #270: [2] Set up caching for specific endpoints.
### Descriptive summary

There may be some endpoints that we will want to cache locally. this allows for the use of controlled vocabs without the worry of an endpoint going down. One we have used in the past was academic affiliation which hit opaquenamespace. This means that we can use all the perks of controlled vocabs from opaquenamespace without the worry of it working. 

The cache should hit the endpoint only when the cache is invalidated. We used the etag to decide whether or not it is invalid. We need to pull the data and cache it locally using rails cache. Then every time after that we check the etags and if they are different, cache the new version, otherwise use the local data. 

If the endpoint is unavailable, use what is cached.

### Expected behavior

When a user types into a controlled field that has a cached piece of data, it utilizes the cached data if it is valid or pulls the new data down, caches, and then uses that instead.

lsat12357:
fixes https://github.com/OregonDigital/hyrax-migrator/issues/270 and grabs other map changes


wickr (approved):
Looks good with most recent MAP pull

PR #1620, opened 2021-04-15
initial configuration to get closer to proper searching

straleyb:


straleyb:
Found an update that doesnt need all these changes

PR #1632, opened 2021-04-20
Feature/add other citations in

straleyb:
Still working on tests and making sure they render proper

### DIFF
--- a/app/views/hyrax/base/_citations.html.erb
+++ b/app/views/hyrax/base/_citations.html.erb
@@ -1,12 +1,12 @@
 <div class="citations">
     <% if Hyrax.config.citations? %>
-      <h3>APA</h3>
+      <h3><%= t('apa_header') %></h3>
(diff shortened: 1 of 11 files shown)


straleyb:
Screen Shot 2021-04-21 at 12 57 07 PM

app/views/hyrax/base/_citations.html.erb
```diff
@@ -1,7 +1,11 @@
 <div class="citations">
     <% if Hyrax.config.citations? %>
+      <h3>APA</h3>
+      <%= export_as_apa_citation(@presenter) %>
       <h3><%= t('.header') %></h3>
```
CGillen, line 5: Lets convert all this to the same method. Either all in partial or all in translations. Translations is probably better.

CGillen (changes requested):

CGillen (approved):

PR #1636, opened 2021-04-22
Upgrade ruby version to 2.7 and switch image base to alpine

CGillen:
@rordway suggested we switch to `alpine` based images and Hyrax now officially supports ruby 2.7. This batches those two upgrades together.


CGillen:
Superseded by #1663

PR #1663, opened 2021-05-25
change from ruby:2.5.8 -> ruby:2.5-alpine3.12

decimalator:


Dockerfile
```diff
@@ -9,26 +9,44 @@ RUN gem install bundler
...
+RUN mkdir -p /tmp/ffi && \
+  curl -sL https://codeload.github.com/libffi/libffi/tar.gz/refs/tags/v3.2.1 \
+  | tar -xz -C /tmp/ffi && cd /tmp/ffi/libffi-3.2.1 && ./autogen.sh &&\
+  ./configure --prefix=/usr/local && make && make install && rm -rf /tmp/ffi
 
 # install FITS for file characterization
 RUN mkdir -p /opt/fits && \
-  curl -fSL -o /opt/fits.zip https://github.com/harvard-lts/fits/releases/download/1.5.0/fits-1.5.0.zip && \
-  cd /opt && unzip fits.zip -d fits/ && chmod +X fits/fits.sh
+  curl -fSL -o /opt/fits-1.0.5.zip http://projects.iq.harvard.edu/files/fits/files/fits-1.0.5.zip && \
```
CGillen, line 41: Why the `fits` regression?

Dockerfile
```diff
@@ -9,26 +9,44 @@ RUN gem install bundler
...
+#  build-essential libpq-dev libreoffice imagemagick graphicsmagick unzip \
+#  zip ghostscript vim tesseract-ocr poppler-utils libopenjp2-tools \
+#  ffmpeg qt5-default libqt5webkit5-dev xvfb xauth openjdk-11-jre \
+#  --fix-missing --allow-unauthenticated
+
+# The alpine way
+RUN apk --no-cache update && apk --no-cache upgrade && \
+  apk add --no-cache alpine-sdk nodejs imagemagick unzip ghostscript vim yarn \
+  git sqlite sqlite-dev postgresql-dev libressl libressl-dev \
+  curl libc6-compat build-base tzdata zip autoconf automake libtool texinfo
```
CGillen, line 26: No java? I'm pretty sure fits requires `java 1.7` for `fits 1.0.5` and `java 1.8` for `fits 1.5.0`
Have you tested to see if fits runs?

CGillen (changes requested):

Dockerfile
decimalator: copied from SA which still uses 1.0.5 direct from Harvard. reverted back to 1.5.0 from github

Dockerfile
decimalator: added the java-common package, and openjdk11-jre-headless to match what we were using before

CGillen (approved):

PR #1683, opened 2021-05-28
Introduce EDTF validation on date fields

### LINKED ISSUE #40: Date terms that adhere to the EDTF Format
### Descriptive summary

There is a format for dates that is compliant with LOC and ISO 8601 called the extended date time format. This format helps format specific date formats, be easier to search for and keeps them consistent. This helps with dates such as `circa 1991` or something along those lines. To read more about EDTF dates, read here https://www.loc.gov/standards/datetime/

We made SA EDTF compliant, in a large PR found here https://github.com/osulp/Scholars-Archive/pull/523 Most of this work will get date fields EDTF compliant. One thing that needs to be decided is which dates out of all the different date types need to be edtf compliant. Do all of them? Some of them? Who knows. 

Porting the work will be quite a process and @luisgreg99 Was the one who championed this. 

### Expected behavior

Dates that are set to follow the edtf format for dates save edit and show pages all show the dates in the proper format.

### Related work

https://github.com/osulp/Scholars-Archive/pull/523

CGillen:
Fixes #40


CGillen:
This adds a `validates` call to all the work models. The tricky part was once the form failed, when it got sent back it HATED that `member_of_collection_ids` was an empty string and not nil. Also, the partial that got updated is aligned with Hyrax 3.0.0 now and was doing the same thing, because the embargo dates have a default value on submission

straleyb (approved):

PR #1684, opened 2021-06-01
ToC Navigation

straleyb:


app/views/hyrax/base/_form_metadata.html.erb
```diff
@@ -4,7 +4,8 @@
   <% end %>
   <% f.object.model_class::ORDERED_TERMS.each do |term|%>
     <% unless f.object.primary_terms.include? term[:name] %>
-      <%= "<a id=\"#{term[:section_name].downcase}\"></a>".html_safe unless term[:section_name].blank? %>
+      <%= "<h1>#{term[:section_name]}</h1>".html_safe unless term[:section_name].blank? %>
+      <%= "<a id=\"#{term[:section_name]}\"></a>".html_safe unless term[:section_name].blank? %>
```
CGillen, line 8: Headers can be hash-linked to, we shouldn't need this line

app/views/layouts/hyrax/dashboard.html.erb
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
+<html lang="<%= I18n.locale.to_s %>" prefix="og:http://ogp.me/ns#">
+  <head>
+    <%= render partial: 'layouts/head_tag_content' %>
+    <%= content_for(:head) %>
+  </head>
+
+  <body data-spy="scroll" data-target=".bs-docs-sidebar" class="dashboard" style="position:relative">
```
CGillen, line 8: We should only need either this (the date-spy, data-target, and optionally data-offset attributes) or the JS.
If you go with JS, you'll have to set this body to `position: relative` some other way (which should be done anyways because inline-styles)

CGillen (changes requested):

app/views/hyrax/base/_form_metadata.html.erb
```diff
@@ -4,7 +4,8 @@
   <% end %>
   <% f.object.model_class::ORDERED_TERMS.each do |term|%>
     <% unless f.object.primary_terms.include? term[:name] %>
-      <%= "<a id=\"#{term[:section_name].downcase}\"></a>".html_safe unless term[:section_name].blank? %>
+      <%= "<h1>#{term[:section_name]}</h1>".html_safe unless term[:section_name].blank? %>
```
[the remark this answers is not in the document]
straleyb: This is an h1 because thats where it starts. I'll have to configure scrollspy further to update it to look at h2 first.

CGillen (approved):

PR #1694, opened 2021-06-10
update appropriate search field text

straleyb:


CGillen (changes requested):
All these at least need a hidden label with the same content. From my research the `title` attribute is not an acceptable substitute. The label should come immediately before the text box and include the `for="<id>"` attribute with the id of the textbox being labeled

CGillen (approved):

PR #1734, opened 2021-07-16
Bulkrax

orangewolf:
This PR adds Bulkrax and all of the OD specific configuration we did.

There are a few artifacts of our dev/test/staging that we can either remove from the PR or you are welcome to if you want them.  These are:

Dockerfile.*. - we run all services in Docker and so we added Dockerfiles for some of the services to preload data on our dev/staging environment. These are used in our override and are just for dev setup / staging setup convienence
bin/firstrun - skipping first run stuff if the admin set exists
chart - we deploy via this helm chart to staging. easy to remove or not. up to you
docker-compose.override.yml-notch8 - dev / test configuration we use
.gitlab and .gitlab-ci.yml - we use gitlab and gitlab pipelines. let me know if you want these removed.


build/firstrun.sh
```diff
@@ -1,5 +1,20 @@
 #!/bin/sh
 
+echo
```
orangewolf, line 3: we found this useful to keep firstrun from happening more than once. mostly for speed

PR #1889, opened 2021-08-12
advanced search title update

straleyb:
Tried this a couple different ways, but this is ultimately what I felt looked the most organized. If we need to extend the functionality of generating page titles, then we can start thinking about generating an object to help.

### DIFF
--- a/app/helpers/hyrax/title_helper.rb
+++ b/app/helpers/hyrax/title_helper.rb
@@ -19,8 +19,8 @@ module Hyrax::TitleHelper
   end
 
   def default_page_title
-    if controller_name.to_s == "advanced"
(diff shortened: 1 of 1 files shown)


CGillen (commented):
This also breaks for regular search.
`index catalog`
What can we do here?

CGillen:
Does removing the `action_name` help more than it hurts? I don't see many pages that actually use the action. I don't know if that's basically useless though, because the BL pages become `advanced` and `catalog`

straleyb:
Screen Shot 2021-08-19 at 9 11 54 AM
Screen Shot 2021-08-19 at 9 12 15 AM

CGillen (approved):

PR #1981, opened 2021-08-24
Update collections/my works/my and admin sets to have table headers per row

straleyb:


straleyb:
Screen Shot 2021-08-24 at 12 53 57 PM

Issue #1877, opened 2021-08-12
[A11y] Explore Collections - Grid - Contrast


jechols:
### Operating System

Windows

### Browser

Edge

### Affected page

Explore Collections

### Admin user?

Yes

### Issue Description

Contrast of white on orange text is failing accessibility checks.

### Suggested Fix

Pick a darker orange?  Or not use orange at all for backgrounds.  Even though black on orange doesn't technically fail WCAG, it's not a great combination.

### Can anybody QA or does this need an expert?

A11y familiarity recommended

### QA Steps

A tool that auto-detects contrast can help here, but some flag text incorrectly (WAVE) while others miss some of the contrast failures (Accessibility Insights).  Most likely this will require a manual check with a color-picker.

CGillen:
Fixed by #1987

PR #1987, opened 2021-08-25
White text on orange background contrast improvements

### LINKED ISSUE #1788: [A11y] Desktop/Mobile Explore Collection - Grid: Orange "All" tab and Active pagination
### Operating System

MacOS

### Browser

Chrome

### Affected page

Explore Collections

### Admin user?

- [ ] Yes

### Issue Description

Explore Collection - Grid: https://staging.oregondigital.org/explore_collections/all?locale=en&view=masonry

Failed Instance:
color-contrast: Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds

Path: .active > .stacker-container > .btn-outline.btn
Snippet: 
`<a class="btn btn-outline" href="/explore_collections/all?locale=en">All</a>`

### Suggested Fix

Fix the following:Element has insufficient color contrast of 2.65 (foreground color: #ffffff, background color: #e28a25, font size: 10.5pt (14px), font weight: normal). Expected contrast ratio of 4.5:1

POSM Review for color options, these color combos pass the color contrast checker:

Orange background and Black Text
orange_black

Tab Example
 orange_black_tab

Blue background and White Text
blue_white

Tab Example
blue_white_tab

Orange and Blue Text, using accessible color generator
AA Compliant
Screen Shot 2021-08-10 at 1 57 14 PM

AAA Compliant
Screen Shot 2021-08-10 at 1 58 43 PM

MAC Palette Styleguide
a55ffd4f-14ff-493e-aa13-ba9ca5fb9208

MAC Palette Styleguide Buttons
55c039ff-37e1-462c-ae2d-2e5f21400afc

If we want to keep orange as a highlight, some options with white and black text

AA compliant, standard
Screen Shot 2021-08-04 at 9 25 20 AM

AAA compliant, strict
Screen Shot 2021-08-04 at 9 25 29 AM

The orange highlight is also used as the active page in the pagination.
Screen Shot 2021-08-05 at 9 27 43 AM

### LINKED ISSUE #1885: [A11y] Explore Collections - Grid - Items per page button
### Operating System

Windows

### Browser

Edge

### Affected page

Explore Collections

### Admin user?

Yes

### Issue Description

The items per page button has a few problems:

- There's no accessible label that explains this since it's just a button.  The "SHOW:" text next to it isn't actually associated programmatically.
- This is basically a disclosure but isn't coded to act like one.  It actually has some menu coding, but it isn't coded as a full menu widget, either.

### Suggested Fix

This could just be a more traditional "Results per page" set of links with no dropdown element.  I'm guessing we want that for the style, but simplifying the UI might be worth consideration.

If not, this needs to be a fully conformant disclosure or menu.  It's probably easier to make a disclosure, but it's so close to being a conformant menu already.  It needs `aria-haspopup="true"` and potentially `aria-controls="IDREF"` and the JS for some of the missing keyboard functionality.

The label for the button needs to reflect the purpose, not the "current value".  It should be "Show X results per page" or similar.  This label could be hidden, but that might be a problem for other WCAG conformance (for instance for low-vision users who use a very high zoom level and a screen reader).

See https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-links.html for an example of a navigation menu button.

Needs expert

### QA Steps

Validation of the menu/disclosure needs expertise to ensure all the keyboard controls work, aria is set properly, etc.

CGillen:
Fixes #1885
Fixes #1788 
#1788 mentions POSM's decision to change the orange we use in cases like this. I added a new `$contrast-orange` to use with `$navy-blue`
When we're all done with a11y tickets, we need to go and audit various links and mouse-over cursors, to make sure enough is changing on hover & on focus to meet a11y standards.

### DIFF
--- a/app/assets/stylesheets/_variables.scss
+++ b/app/assets/stylesheets/_variables.scss
@@ -2,6 +2,7 @@
 $navy-blue: #073C5C;
 $apple-green: #89BC41;
 $lime-green: #D0DD28;
+$contrast-orange: #E88F2A;
 $dark-orange: #E28A25;
 $light-orange: #FDBB49;
 
--- a/app/assets/stylesheets/oregon_digital/_search.scss
+++ b/app/assets/stylesheets/oregon_digital/_search.scss
@@ -9,8 +9,8 @@
   border: none;
   box-shadow: none;
   &.active {
-    color: white;
-    background-color: $dark-orange;
+    color: $navy-blue;
+    background-color: $contrast-orange;
   }
 }
 
@@ -210,8 +210,9 @@ div.pagination {
 }
 ul.pagination li {
   &.active span {
-    background-color: $dark-orange;
+    background-color: $contrast-orange;
     border: none;
+    color: $navy-blue;
   }
   &.next span,
   &.next a,
--- a/app/assets/stylesheets/oregon_digital/_search.scss
+++ b/app/assets/stylesheets/oregon_digital/_search.scss
@@ -220,6 +220,16 @@ ul.pagination li {
   &.previous a {
     background: none;
   }
+  &.disabled {
+    span,
+    span:hover,
+    span:focus,
+    a,
+    a:hover,
+    a:focus {
+      color: $dark-grey;
+    }
+  }
 
   a,
   span {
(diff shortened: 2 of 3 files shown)


straleyb (approved):

CGillen:
Does not fix #1885, actually fixes #1877

PR #2029, opened 2021-09-01
Ticket #1746: Commit to add aria-current and rework link, so that cur…

### LINKED ISSUE #1746: [A11y] Desktop/Mobile Collection Landing: Breadcrumbs Active Page
### Operating System

MacOS

### Browser

Chrome

### Affected page

Collection Show

### Admin user?

- [ ] Yes

### Issue Description

Collection Landing Page: https://staging.oregondigital.org/collections/gwilliams?locale=en

When I tab through the page and land on the breadcrumbs it will tab and read "Home" but skips and doesn't read "Geralds W. Williams Collection" the active page I am on.

### Suggested Fix

I mentioned in a previous accessibility ticket #1745 for the possible fix.

a element needs the following:
aria-current=page

See w3c recommendation:
https://w3c.github.io/aria-practices/examples/breadcrumb/index.html

thornka:
…rent page of breadcrumbs is recognized (fixes #1746)


app/controllers/concerns/hyrax/breadcrumbs_for_collections.rb
```diff
@@ -0,0 +1,29 @@
+# OVERRIDE FROM HYRAX to add aria-current tag
```
CGillen, line 1: I don't think this override is necessary anymore

app/builders/hyrax/bootstrap_breadcrumbs_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+  
+      @context.tag.nav(breadcrumbs_options) do
+        @context.tag.ol do
+          safe_join(@elements.uniq.collect { |e| render_element(e) })
+        end
+      end
+    end
+  
+    def render_element(element)
+      html_class = 'active' if @context.current_page?(compute_path(element))
```
CGillen, line 24: You might not need to, but lets leave our overrides as similar to Hyrax as we can. Revert this change

app/builders/hyrax/bootstrap_breadcrumbs_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+
```
CGillen, line 25: Add a new line here
`element.options['aria-current'] = 'page' if html_class`. This will help us all the way across the app =]

CGillen (changes requested):

app/builders/hyrax/bootstrap_breadcrumbs_builder.rb
CGillen: Feel free to push back if you're seeing or testing otherwise

CGillen (approved):

PR #2048, opened 2021-09-09
Ticket #1870: Commit to change accessible name for i icons (fixes #1870)

### LINKED ISSUE #1870: [A11y] Collection Landing - Link purpose - "i" icon
### Operating System

Windows

### Browser

Edge

### Affected page

Collection Show

### Admin user?

Yes

### Issue Description

The "i" next to "Oregon State University" links to LC's page about the term, but for some reason its accessible name is "Open link in new window".  This one is especially confusing because an aria label is used to override the attempt at making that link text be "learn more", but that also isn't a meaningful piece of text.

### Suggested Fix

The "i" icons need some kind of explanation about where they take you, and even if "learn more" were the accessible name, that would be iffy.  In most cases the link text can be the destination page's title, so in this case it could be "Oregon State University - LC Linked Data Service...".  But this icon is probably a pretty general thing we use, so a more generic term might be good like "learn more about this taxonomy term" or something.  Dunno.  This should probably get more experts in the discussion.

Needs expert

### QA Steps

Verify that the accessible name of the icon is meaningful within the context of the item.  Exact verbiage requires discussion, and I'll update this section once we figure that out, but testing that it's at least better than "Open link in new window" should be pretty easy :)

thornka:


CGillen (changes requested):
Looks great, all you gotta do is fix your spec
https://github.com/OregonDigital/OD2/blob/master/spec/renderers/hyrax/renderers/search_and_external_link_attribute_renderer_spec.rb

CGillen (approved):

Issue #1868, opened 2021-08-11
[A11y] Explore Collections - Grid - Parsing


jechols:
### Operating System

Windows

### Browser

Edge

### Affected page

Explore Collections

### Admin user?

Yes

### Issue Description

Duplicated id "explore-collection-search"

### Suggested Fix

Fix either the input's id or the form's id

### Can anybody QA or does this need an expert?

A11y familiarity recommended

### QA Steps

Verify (looking at HTML or using something like the WCAG DOM checker) that there are no duplicated IDs

jechols:
So the issue originally described here is good, but now there's a new duplicated id, `per_page-dropdown`. I'm going to leave this ticket open since it's the same problem, but if you want me to, I can make a new ticket.

Issue #1893, opened 2021-08-12
[A11y] Desktop/Mobile Explore Collection Grid: Nu Checker


azle:
### Operating System

MacOS

### Browser

Chrome

### Affected page

Explore Collections

### Admin user?

No

### Issue Description

Explore Collection Grid: https://staging.oregondigital.org/explore_collections/all?locale=en&view=masonry

Below are the Nu Checker Errors:
[ExploreCollectionGrid.zip](https://github.com/OregonDigital/OD2/files/6978070/ExploreCollectionGrid.zip)

### Can anybody QA or does this need an expert?

Needs expert

PR #2121, opened 2021-12-09
Ticket #1868: Commit to fix duplicated per_page-dropdown ids (fixes #…

thornka:
…1868)


CGillen:
Oh dang, sorry, I just fixed this from this ticket: #1893 Which included the double id issue in the list of invalid HTML. We do actually need that extra set of `results_collection_tools` because in mobile vs desktop they're too far apart and need different CSS.

PR #2132, opened 2022-01-05
Update headings on homepage and adjust css

### LINKED ISSUE #1906: [A11y] Homepage - Headings
### Operating System

Windows

### Browser

Edge

### Affected page

Homepage

### Admin user?

Yes

### Issue Description

- There are two level one headings and they're both the same, which is confusing
- The level 2 headings make sense, though they shouldn't be all-caps (e.g., "FEATURED COLLECTIONS").  If it has to be all-caps for stylistic reasons, it should use CSS for that rather than be written out as all-caps.  When CSS transforms something into all-caps, user agents can "undo" the transformation and screen readers know what the real text is.
- The collection names shouldn't be headings I don't think, because they aren't really above a section of content.  I could be wrong, though, because maybe there's the potential for content under these headings and our data just happens to not have any.
- The "X ITEMS   VIEW ALL" bits are level four headings and definitely shouldn't be coded this way

### Suggested Fix

- Remove one of the level one headings.  It is preferable to leave a level one heading near the main content, so the logo probably should just not be a heading at all.
- Change level 2 headings to normal case (e.g., "FEATURED COLLECTIONS" to "Featured Collections") and transform via CSS.
- Consider taking away the level 3 headings around the collection titles.  Also consider leaving them there.  Very subjective.
- Remove the `<h4>` around the item count / view all.

A11y familiarity recommended

### QA Steps

Check the page heading structure, preferably with a tool that can outline the headings on the page so it's clear which ones are where on the page.  Make sure the headings seem to present an outline of the page content.

straleyb:
Fixes #1906 

The duplicate h1 had been removed already but the adjustments to the h2 and h4 elements has been made.

### DIFF
--- a/app/assets/stylesheets/oregon_digital/_home-page.scss
+++ b/app/assets/stylesheets/oregon_digital/_home-page.scss
@@ -216,7 +216,3 @@ nav.navbar-default {
   margin-left: 100px;
   padding-right: 140px;
 }
-
-.upper-case {
-  text-transform: uppercase;
-}
--- a/app/assets/stylesheets/oregon_digital/_home-page.scss
+++ b/app/assets/stylesheets/oregon_digital/_home-page.scss
@@ -216,3 +216,7 @@ nav.navbar-default {
   margin-left: 100px;
   padding-right: 140px;
 }
+
+.upper-case {
+  text-transform: uppercase;
+}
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/oregon_digital/_home-page.scss
```diff
@@ -215,4 +215,8 @@ nav.navbar-default {
 .about-oregondigital-container {
   margin-left: 100px;
   padding-right: 140px;
+}
+
+.upper-case {
```
CGillen, line 220: We already have `.uppercase` and `.lowercase`
in `fonts.scss`

CGillen (changes requested):

CGillen (approved):

PR #2163, opened 2022-01-18
update the h2 to just be a div with h2 class

### LINKED ISSUE #2002: [A11y] Work Show Page-All - Headings
### Operating System

Windows

### Browser

Edge

### Affected page

Work Show

### Admin user?

Yes

### Issue Description

https://staging.oregondigital.org/concern/generics/df727084j
https://staging.oregondigital.org/concern/documents/df727057n
https://staging.oregondigital.org/concern/images/1g05fb620

There are two headings that don't make sense, and are only causing problems for screen-reader users, so they can't be considered usability problems. "SEARCH CONSTRAINTS" and "SEARCH RESULTS" are both showing up above various other asset lists ("CONTAINED IN THIS OBJECT" for compound parents, "IS PART OF" and "RELATED ITEMS" on children - this doesn't affect single works). This makes three headings for one area, two of which have no content, and the one that's valid is actually the furthest away from its content.

Additionally, the contact form on all work pages has a level 1 heading, "Contact Form". This is both at the wrong level (it should be a level 3 heading since the tab is a level 2 heading) and unnecessary since the tab is already a heading.

### Suggested Fix

- Remove the extraneous headings related to search: "SEARCH CONSTRAINTS" and "SEARCH RESULTS"
- Remove the level 1 heading above the contact form or else change it to a level 3 heading

A11y familiarity recommended

### QA Steps

Accessibility Insights or potentially a tool like Webaim's WAVE validator can help find hidden headings. Verify heading structure makes sense, there are no empty headings (headings with no content under them), and that the heading names are meaningful.

The level 1 heading above the contact form is only visible when "CONTACT US" is the active tab, and Accessibility Insights has trouble with that. Consider a different tool or multiple tools for verifying correctness of headings.

straleyb:
Fixes #2002 

This fixes the h2 issue, the contact form on the show page already has the h3

### DIFF
--- a/app/views/hyrax/base/_search_results.html.erb
+++ b/app/views/hyrax/base/_search_results.html.erb
@@ -1,5 +1,3 @@
-<div class="h2 sr-only top-content-title"><%= t('blacklight.search.search_results_header') %></div>
-
 <% @page_title = t('blacklight.search.page_title.title', :constraints => render_search_to_page_title(params), :application_name => application_name) %>
 
 <% content_for(:head) do -%>
@@ -7,8 +5,6 @@
 <% end %>
 
 <%= render 'search_header' %>
-
-<div class="h2 sr-only"><%= t('blacklight.search.search_results') %></div>
 
 <%- if @response.empty? %>
   <%= render "zero_results" %>
--- a/app/views/hyrax/base/_search_results.html.erb
+++ b/app/views/hyrax/base/_search_results.html.erb
@@ -1,4 +1,4 @@
-<h2 class="sr-only top-content-title"><%= t('blacklight.search.search_results_header') %></h2>
+<div class="h2 sr-only top-content-title"><%= t('blacklight.search.search_results_header') %></div>
 
 <% @page_title = t('blacklight.search.page_title.title', :constraints => render_search_to_page_title(params), :application_name => application_name) %>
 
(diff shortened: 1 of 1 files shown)


app/views/hyrax/base/_search_results.html.erb
```diff
@@ -1,4 +1,4 @@
-<h2 class="sr-only top-content-title"><%= t('blacklight.search.search_results_header') %></h2>
+<div class="h2 sr-only top-content-title"><%= t('blacklight.search.search_results_header') %></div>
```
CGillen, line 1: I don't think either of these lines are necessary. We don't use search constraints here.

app/views/hyrax/base/_search_results.html.erb
```diff
@@ -8,7 +8,7 @@
 
 <%= render 'search_header' %>
 
-<h2 class="sr-only"><%= t('blacklight.search.search_results') %></h2>
+<div class="h2 sr-only"><%= t('blacklight.search.search_results') %></div>
```
CGillen, line 11: And we don't need to label these as search results, they should just look like a list of works

CGillen (changes requested):

CGillen (approved):

PR #2184, opened 2022-02-01
Use autocomplete attribute on contact forms

### LINKED ISSUE #2024: [A11y] Work Show Page-All - Autofill in contact form
### Operating System

Windows

### Browser

Edge

### Affected page

Work Show

### Admin user?

Yes

### Issue Description

https://staging.oregondigital.org/concern/generics/df727084j
https://staging.oregondigital.org/concern/documents/df727057n
https://staging.oregondigital.org/concern/images/1g05fb620

The contact form has two text fields which have a clear meaning, but aren't coded properly for autofill / programmatic input purpose.

See:

- https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html
- https://www.w3.org/TR/WCAG21/#input-purposes
- https://www.w3.org/TR/html52/sec-forms.html#sec-autofill

### Suggested Fix

Use the proper autofill attribute to describe the input's purpose: I believe `autofill="name"` and `autofill="home email"` make the most sense per the spec.

Needs expert

### QA Steps

Verify the autofill value is correct in describing the purpose and in how the spec defines valid autofill attribute values. It gets complicated to unravel: https://www.w3.org/TR/html52/sec-forms.html#autofilling-form-controls-the-autocomplete-attribute

CGillen:
Fixes #2024 
Note: This only works on browsers that support the new HTML 5 values for `autocomplete`. At time of writing that is Chrome & Opera. This is not a breaking attribute for other browsers and they should implement this standard over time.
Because this is "new" it is not supported by Firefox or Safari yet.

### DIFF
--- a/app/views/hyrax/base/_work_contact_form.html.erb
+++ b/app/views/hyrax/base/_work_contact_form.html.erb
@@ -23,12 +23,12 @@
 
   <div class="form-group">
     <%= f.label :name, t('hyrax.contact_form.name_label'), class: "col-sm-2 control-label" %>
-    <div class="col-sm-10"><%= f.text_field :name, value: nm, class: 'form-control', required: true %></div>
+    <div class="col-sm-10"><%= f.text_field :name, value: nm, class: 'form-control', autocomplete: 'given-name', required: true %></div>
   </div>
 
   <div class="form-group">
     <%= f.label :email, t('hyrax.contact_form.email_label'), class: "col-sm-2 control-label" %>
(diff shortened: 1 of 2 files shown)


jechols:
FWIW, autocomplete isn't really an attribute browsers have to do anything with, so if it doesn't work in FF that's not a deal-breaker. It's designed as a way to *allow* the user agent to help with the field more than as a part of what any given browser is expected or required to do. e.g., an add-on could overlay icons to help people who have dyslexia to know exactly which fields mean what.

straleyb (approved):

PR #2193, opened 2022-02-03
Ticket #1909: Commit to fix nested divs within label (fixes #1909)

### LINKED ISSUE #1909: [A11y] Homepage - Parsing error
### Operating System

Windows

### Browser

Edge

### Affected page

Homepage

### Admin user?

Yes

### Issue Description

There are a lot of nested divs inside labels, which is a parsing failure.  I can't find the actual control on the page that makes this show up, so I'm thinking this HTML might actually be a mistaken widget that doesn't need to be here in the first place (though if it's in use elsewhere it would still need to be fixed).

Here's an example of the HTML:

```html
          <div class="select-worktype">
            <label>
              <input type="radio" name="payload_concern" value="Image" data-single="/concern/images/new?locale=en" data-batch="/batch_uploads/new?locale=en&amp;payload_concern=Image">
              <div class="select-work-icon">
                <span class="fa fa-file-text-o"></span>
              </div>
              <div class="select-work-description">
                <h3 class="work-type-title">Image</h3>
                <p>Image works</p>
              </div>
            </label>
          </div>
```

### Suggested Fix

Just make the `<p>Image works</p>` into a label and associate it with the radio button normally (using the `for` attribute).

The label seems wrong in this case anyway, and should just be wrapping the "Image works" bit, I think.  Wrapping the whole thing means the label for that radio button is "Image Image works", which makes less sense.

A11y familiarity recommended

### QA Steps

The WCAG scan will need to be used to verify nothing is broken on this page.  This is most easily done with the Accessibility Insights bookmarklets.

Source will need to be inspected to verify the label is properly associated, since that will change when fixing the invalid nesting.

thornka:


CGillen (changes requested):
I think we're still going to need `row_presenter.name`. IDK if it still needs to be `h3` or not (I haven't gone through the dashboard a11y tickets yet), but it should still be displayed.
The simplest solution is to just change all the `div`s from the original file to `span`s
I trust whichever solution you choose, but we still need the name of the work type

CGillen (approved):

PR #2195, opened 2022-02-03
uses em without background highlighting

### LINKED ISSUE #1914: [A11y] Homepage - Semantics - Emphasis
### Operating System

Windows

### Browser

Edge

### Affected page

Homepage

### Admin user?

Yes

### Issue Description

The following text is visually emphasized:

> A collaboration between the University of Oregon Libraries and Oregon State University Libraries & Press, Oregon Digital comprises more than 100,000 digitized works including historic and modern photographs, manuscripts, and publications.

It seems this is emphasized by the font, not CSS and not a semantic tag.  This makes it extra tricky, because if we add an `<em>` tag, the text will change to have a light blue background for some reason.  But if we don't add that tag, we're offering a visual emphasis where there is no semantic emphasis.

### Suggested Fix

Use a standard font and the `<em>` tag if possible.  Otherwise this probably needs to go to somebody with more expertise to figure out how important it is to make sure emphasis is properly tagged.  I've heard anecdotes that this isn't that important, but that doesn't mean it's not a hard failure - I'm just not sure.

A11y familiarity recommended

### QA Steps

You have to verify that the markup is correct, not just the visual display, so this needs at least some familiarity to check that `<em>` is in use.

straleyb:
Fixes #1914

### DIFF
--- a/app/assets/stylesheets/oregon_digital/_home-page.scss
+++ b/app/assets/stylesheets/oregon_digital/_home-page.scss
@@ -223,3 +223,7 @@ ul.nav.homepage-nav {
   margin-left: 100px;
   padding-right: 140px;
 }
+
+.semantic-em {
+  background-color: $white;
+}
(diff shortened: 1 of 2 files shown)


CGillen:
See notes on #1914 about why `<em>` and `<i>` don't make sense or are needed here

PR #2196, opened 2022-02-03
Ticket #1910: Commit to remove title from masthead image (fixes #1910)

### LINKED ISSUE #1910: [A11y] Homepage - Images
### Operating System

Windows

### Browser

Edge

### Affected page

Homepage

### Admin user?

Yes

### Issue Description

The alt attribute for the background image on the homepage is "masthead image".  This image seems to me to be purely decorative, as I can remove it and the page seems exactly as usable as with it.

### Suggested Fix

If this image is decorative, its alt attribute must be null (`alt=""`).

If it's not decorative, its alt attribute needs to be more useful.  I don't know what it would be that would explain its function, however, as its function seems to be purely decoration to me.

A11y familiarity recommended

### QA Steps

Verify the alt attribute makes sense: null if decoration, short but meaningful otherwise.

thornka:


thornka:
Fixes #1910

CGillen (approved):

PR #2201, opened 2022-02-08
Ticket #1908: Commit to set focus to search content

### LINKED ISSUE #1908: [A11y] Homepage - Status Messages
### Operating System

Windows

### Browser

Edge

### Affected page

Homepage

### Admin user?

Yes

### Issue Description

Search results from the homepage doesn't automatically identify anything about the status of the search.  The search pages are always titled "Index Catalog" and there's nothing that programmatically helps users "see" the results.

### Suggested Fix

- On no results, the title should probably be "Search: No Results" or similar
- On a search that has results the title could say something like "Search Results"
- The focus could be automatically taken directly to the results area to assist screen-reader and keyboard users.

A11y familiarity recommended

### QA Steps

The title could be tested by anybody.  If focus is changed programmatically, that would require a keyboard user and possibly a screen reader to test that the focus is in a useful place after page load.

### LINKED ISSUE #1909: [A11y] Homepage - Parsing error

thornka:
(fixes #1908  )

@CGillen confirmed that the title issue had been previously fixed. 

I chose to automatically place the focus on the view type group because that seemed like the first piece of search content. Let me know if I need to place the focus on a different element or if you want the <script> in a separate file.

### DIFF
--- a/app/assets/javascripts/hyrax/search.js
+++ b/app/assets/javascripts/hyrax/search.js
@@ -69,3 +69,7 @@
 Blacklight.onLoad(function() {
   $('#search-form-header').search();
 });
+
+Blacklight.onLoad(function() {
+  location.hash = 'content'
+});
--- a/app/views/catalog/_constraints.html.erb
+++ b/app/views/catalog/_constraints.html.erb
@@ -4,7 +4,7 @@
   <%= render_constraints(params) %>
 </div>
 <div class="widget-wrapper">
-  <div class="pull-right catalog-search">
+  <div id="catalog-search" tabindex="0" class="pull-right catalog-search">
     <%# OVERRIDE FROM BLACKLIGHT to remove start over link and replace with view type widget %>
     <% tool_config = Blacklight::Configuration::ToolConfig.new %>
     <% tool_config.name = :view_type_group %>
@@ -21,3 +21,7 @@
 </div>
 
 <div class="constraints-divider"></div>
+
+<script>
+  document.getElementById('catalog-search').focus()
+</script>
--- a/app/views/catalog/_constraints.html.erb
+++ b/app/views/catalog/_constraints.html.erb
@@ -4,7 +4,7 @@
   <%= render_constraints(params) %>
 </div>
 <div class="widget-wrapper">
-  <div id="catalog-search" tabindex="0" class="pull-right catalog-search">
+  <div class="pull-right catalog-search">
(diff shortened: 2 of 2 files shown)


CGillen (changes requested):
Yup, you got it, we'll want the script to be in it's own or an existing JS file. Minimize `<script>` tags. Ideally only the `javascript_include_tag` in `views/layout/*`
You can probably find an existing JS file that loads only with blacklight. Like `hyrax/search.js`
In there you can do something a little more simple, like set the browser's location hash to an existing (or new) element with an id (no focus necessary). We also want to skip the results and skip the constraints. That _might_ end up a little confusing for users, but the ticket specifically asks for the results section.
So, some JS like this might be enough:
```
Blacklight.onLoad(function() {
  location.hash = 'content'
});
```

thornka:
Adding the JS into hyrax/search.js seemed to work well for keyboard users, but I'm unsure about the functionality for screen readers

CGillen (approved):

PR #2203, opened 2022-02-08
Ticket #2188: Commit to fix tabindex on sign up and thumbnails

### LINKED ISSUE #2188: [A11y] Home - Items unreachable via tab key
### Operating System

Windows

### Browser

Firefox

### Affected page

Homepage

### Admin user?

No

### Issue Description

Various items are not in the tab sequence. That is to say, they're not focusable via the tab key (or if they are, I'm not seeing how to get to them).

- The "Sign Up" link at the top, next to the "LOG IN" button, seems to have an explicit tab index of -1, which means it cannot be reached except with JavaScript
- The collections' thumbnails also have a tab index of -1

### Suggested Fix

I think just removing the tab index attribute would fix this.

Anybody

### QA Steps

- Using the tab key alone, make sure you can navigate to "Sign In" and activate it (via the spacebar or enter key)
- Using the tab key alone, make sure you can navigate to collections' featured images and activate them (these are links, so only the enter key needs to work)

thornka:
(fixes #2188 )

### DIFF
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -33,8 +33,7 @@
     </li>
     <li>
       <%= link_to t('hyrax.dashboard.my.action.sign_up'), main_app.new_user_registration_path,
-                   class: 'btn btn-primary login-button',
-                   tabindex: -1 %>
+                   class: 'btn btn-primary login-button' %>
     </li>
     <!--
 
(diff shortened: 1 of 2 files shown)


CGillen (changes requested):
Looks great, just remove `#5feae087457577878dd74cf504d4b70dbfeee816`

CGillen (approved):

PR #2237, opened 2022-03-17
[WIP] start a prometheus exporter endpoint service for workers

decimalator:


decimalator:
adds a method to start the prometheus exporter for Sidekiq workers

PR #2283, opened 2022-04-14
Feature/alphabetize facets within hits

straleyb:


db/schema.rb
```diff
@@ -325,6 +325,22 @@
     t.index ["uri"], name: "index_qa_local_authority_entries_on_uri", unique: true
   end
 
+  create_table "qa_mesh_trees", id: :serial, force: :cascade do |t|
```
CGillen, line 328: Some testing data?

CGillen (changes requested):

db/schema.rb
straleyb: Not sure, but any time I blow away the entire thing, rebuild and reset, it asks me to run these migrations. So I figured I'd push them up finally. Its just part of the build from what Ive seen.

CGillen: Weird, It's not coming up for me. Lines 56/57 are though.
Did you do something with Questioning Authority?

straleyb: Nope. I just do a `dcd -v` and then do a `dcb` to rebuild. After I get it set up and start running rake tasks it tells me that there are migrations pending, then i run the migrations and this is what I get. I can try it one more time locally to see if I get the same result.

PR #2394, opened 2022-07-06
set reindex_extent for colls

### LINKED ISSUE #2387: set reindex_extent with an environment variable
### Descriptive summary
hyrax uses reindex_extent to improve  performance by preventing an entire coll from reindexing when it isn't necessary
migrating joel-palmer cpds ended up being unexpectedly painful, as reindex_extent is not set in the case of attaching children to the parent. For each cpd, the entire 5k+ coll was reindexed.
Tamsin Johnson confirms that we can add a method to collection.rb to always limit reindexing, and by using an environment variable, and setting it on just the workers assigned to migration, we can limit the scope
He also suggested we go ahead and do a full reindex, once the coll is complete

### Related work

### Accessibility Concerns

lsat12357:
fixes #2387 
(also adds label exists verifier to list of default verifiers in the migrator config)

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -38,6 +38,10 @@ class Collection < ActiveFedora::Base
     end
   end
 
+  def reindex_extent
+    OD2::Application.config.reindex_extent
+  end
+
   private
 
   # Build new Facet objects that might not exist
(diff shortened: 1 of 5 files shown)


lsat12357:
for this to work as expected, will need to actually set the environment variable on the migration workers.

wickr (approved):

PR #2453, opened 2022-08-12
updates so users can edit other people's names

straleyb:
Enables user information to be edited by an admin

### DIFF
--- a/app/jobs/user_edit_profile_event_job.rb
+++ b/app/jobs/user_edit_profile_event_job.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-
 # Log user profile edits to activity streams
 class UserEditProfileEventJob < EventJob
 	def perform(user)
--- a/app/jobs/user_edit_profile_event_job.rb
+++ b/app/jobs/user_edit_profile_event_job.rb
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 # Log user profile edits to activity streams
 class UserEditProfileEventJob < EventJob
-    def perform(user)
-      @user = user
-      super
-    end
-      
-    def action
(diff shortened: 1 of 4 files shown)


app/jobs/user_edit_profile_event_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+# Log user profile edits to activity streams
+class UserEditProfileEventJob < EventJob
+    def perform
+      super
+    end
+      
+    def action
+      "User profile has been updated by #{current_user}"
```
CGillen: This is performed async, there shouldn't be a `current_user`

app/jobs/user_edit_profile_event_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Log user profile edits to activity streams
+class UserEditProfileEventJob < EventJob
+  def perform
+    super
+  end
+
+  def action
+    "User profile has been updated by #{current_user}"
```
CGillen, line 10: This is performed async, there shouldn't be a `current_user`

CGillen (changes requested):
https://github.com/OregonDigital/OD2/blob/79dfec146e304653722ae254c8536f906de6b30f/app/controllers/hyrax/dashboard/profiles_controller.rb#L60
Won't this blow up the trophies of the editor, not the user?

app/jobs/user_edit_profile_event_job.rb
CGillen: Might want to add an arg to `#perform` for the editor

CGillen (approved):
Approved for squash & merge

Issue #2083, opened 2021-10-20
Configure Date facet on advanced search to use Range slider


straleyb:
### Descriptive summary

Currently date is just a drop down of dates. This is the easiest and most basic usage for the date facet on the advanced search page. We need to try and figure out how to render the date slider on the advanced search page. This is usually done using `range: true` on the facet configuration in the catalog controller. Since this is a unique subset of facets, we cant render them regularly. We need to find out how to render the date range facet manually.

straleyb:
Need to come up with something super special

CGillen:
Here's the ticket in blacklight_range_limit:
https://github.com/projectblacklight/blacklight_range_limit/issues/58

CGillen:
@jsimic Before I spend any more time on this, is this still a requirement? The wireframes has `decade` as a type-ahead now:
https://drive.google.com/drive/u/1/folders/1bF30M2ftI50Syhhg7hL4MIOtn9DmZFEj

jsimic:
@CGillen The type ahead is not useful if you are looking for a large range of dates, which would require a user to select all the individual years within the range. Could we use a two box (earliest and latest) strategy?

CGillen:
I can do an earliest/latest type ahead pair. My biggest issue was the histogram and sliders not being compatible with advanced search.
I did also manage to get decade working, but that's still not a huge range

jsimic:
@CGillen Please change the "-" between the date boxes to the word "to" and center it horizontally and vertically.

jsimic:
QA pass

PR #2572, opened 2022-11-02
Simplify decade indexing and account for negative values

CGillen:
Change advanced search date facet to decade (inline with wireframe) & fix decade indexing for negative years.
Kind of, sort of, maybe related to #2083


CGillen:
Far to old and out-of-date to be useful

PR #2580, opened 2022-11-09
uses dynamic form generation for creation/update of collection w/ custom id

straleyb:
On create the id can be customized, on update id is removed from form. Ids cannot be updated.

ID is required


CGillen (approved):
`#terms` and `#primary_terms` need their tests fixed.

PR #2590, opened 2022-11-15
comment out possible unnecessary perms

straleyb:
@CGillen Wanna check this out? This is the fix i found for the logged in users not being able to generate collections. The controllers were blocking the access through rails stuff and bypassing our cancan permissions. The other idea would be to figure out how to refine the DashboardAuthorization to be able to check if they are trying to create a collection or create a new authorization. Im not sure we need it though.

Thoughts?

### DIFF
--- a/app/controllers/concerns/hyrax/dashboard_authorization.rb
+++ b/app/controllers/concerns/hyrax/dashboard_authorization.rb
@@ -14,10 +14,12 @@ module Hyrax
       render(file: File.join("public/#{response_code}.html"), status: response_code, layout: false) unless dashboard_allowed
     end
 
+    #rubocop:disable Metrics/CyclomaticComplexity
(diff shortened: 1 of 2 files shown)


CGillen:
Do users need both controllers?
Least privilege principle says we should have a check for `correct controller && correct action && correct collection type`. If it's not too tangled in the weeds, can you add a `#is_creating_user_collection` check to `Hyrax::DashboardAuthorization`?

straleyb:
They do need both controllers in order to create a collection. I disabled all the Dashboard Authorizations and slowly turned each one on until I found the ones that were breaking the system. I could upload with all of them turned on besides these two. I can try and add in some extra checks to see what controller we are in and whether or not we are trying to create a collection.

CGillen (approved):

PR #2605, opened 2022-11-22
updates position and size of text box and share buttons

### LINKED ISSUE #1313: Better user feedback for link copying
### Descriptive summary
People might not be able to easily recognize an action occurred when they click the current copy URL links. We should improve the feedback with toasts and a text box for the paranoid.

### Expected behavior
User can see URL they are about to copy in an input box.
Input box auto-selects all text when entered into.
Clickable 'copy URL' button adjacent to box.
Toast or other clear messaging when 'copy URL' button is clicked. 

### Related work
#1312 

### Accessibility Concerns
The current JS copy library doesn't do a great job at using things like the `title` field. Make sure `title` and other HTML components are used properly.

straleyb:
Fixes #1313

### DIFF
--- a/app/views/hyrax/base/_social_media.html.erb
+++ b/app/views/hyrax/base/_social_media.html.erb
@@ -4,7 +4,11 @@
   <span class="mw-lg-b">
     Share this:
   </span>
-  <!-- Sharingbutton Facebook -->
+  <label class="sr-only" for="lname">Copy Link</label>
+  <input class="copy-input" type="text" id="lname" name="lname" value=<%= "#{share_url}" %> style="width: 550px;" onClick="this.select();">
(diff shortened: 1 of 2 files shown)


app/views/hyrax/base/_social_media.html.erb
```diff
@@ -4,7 +4,11 @@
   <span class="mw-lg-b">
     Share this:
   </span>
-  <!-- Sharingbutton Facebook -->
+  <label class="sr-only" for="lname">Copy Link</label>
+  <input class="copy-input" type="text" id="lname" name="lname" value=<%= "#{share_url}" %> style="width: 550px;" onClick="this.select();">
```
CGillen, line 8: Remove `style` to apply CSS

CGillen (changes requested):

CGillen (approved):

Issue #2049, opened 2021-09-09
Add text about what user gets for signing up


CGillen:
### Descriptive summary
Add a short blurb about what the user gets for signing up @kmthorn @petersec for text.
Add right below `Sign Up` bold text.

### Expected behavior
Short paragraph below sign up text

kmthorn:
Add the following text to the right of the sign up form.

# Looking to Sign Up for an Oregon Digital Account?

## I am affiliated with UO or OSU.

- Great! You do not need to sign up for an account. UO and OSU affiliates can just login with your university credentials.
- Once done, you will have access to all of the material that your university offers through Oregon Digital, as well as the ability to select, organize, and share Oregon Digital materials with others.

## I am not affiliated with UO or OSU.

- If you are not affiliated then you need to sign up for an account
- Once done, you will have the ability to select, organize, and share Oregon Digital materials with others.

straleyb:
To right of sign up.

straleyb:
QA: This will show up next to the sign up form on the sign up page. An unregistered user visits the site, clicks sign up, to the right of the sign up form shows the information displayed above.

straleyb:
This is ready for QA. Its related to #2049 and has been looked at by chris and azle.

petersec:
QA Pass (ref #2609 )

PR #2609, opened 2022-11-23
Updates view to include why sign up

straleyb:
Screen Shot 2022-11-23 at 10 57 19 AM

### DIFF
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -1,10 +1,35 @@
-<div class="col-md-3">
-</div>
-<div class="devise-container col-md-6">
-  <h1 class="h2 text-center" style="margin-bottom:2em">Sign up for Oregon Digital</h1>
+<div class="devise-container">
+  <h1 class="h2 text-center" style="margin-bottom:2em">LOOKING TO SIGN UP FOR AN OREGON DIGITAL ACCOUNT?</h1>
+  <%- if controller_name != 'sessions' %>
+    <div class="text-center" style="margin-bottom:2em">Already have an account? <%= link_to "Log in", new_session_path(resource_name) %><br /></div>
+  <% end -%>
+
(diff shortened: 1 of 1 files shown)


CGillen:
@azle Check images above against issue #2049? Is this the direction we should go? Contact @kmthorn (if possible) about continuing to include the sidebar contents

azle:
@CGillen and @straleyb I talked it over with Julia and we wanted to reformat the following page to be more consistent with the other login pages. I feel like the bulleted list is not necessary the user should know what they are signing up for already. If the bulleted list is needed. I prefer for it to be a tooltip using a question mark icon in the same line as the I AM AFFILIATED OR NOT AFFILIATED. Let me know if we need to talk about the changes.

This is what we would like it to look like.

 Screen Shot 2022-12-01 at 10 24 41 AM

straleyb:
@kmthorn Are we good with moving forward in this way?

straleyb:
@kmthorn Pinging again and adding @petersec  to help try and resolve this question about how we want to display the why sign up information or if we want to.

petersec:
@straleyb @CGillen @azle @kmthorn @jsimic - For me, the salient text from the old bulleted list is "select, organize and share Oregon Digital materials with others." As long as that's available somewhere (or somehow) on the page, then I think we've conveyed the information that we need to the user. 

I'd be fine with the question mark and tooltip approach that Azle suggested. Perhaps the tooltip language could read: "By signing up or signing in, you will have the ability to select, organize and share Oregon Digital materials with others."

jsimic:
Chris' wording is great!

straleyb:
Screen Shot 2023-01-04 at 2 08 41 PM

straleyb:
@azle I posted a screenshot of what I got set up currently. Is there anything on there that needs to be changed?

azle:
@straleyb, can we make some adjustments with the line break for the following header "Looking to sign up..." see screen snap.

Can you also change the following text "Affiliated..." and "Not Affiliated..." to the same text as the screen snap.

Also for the links below the input forms do we need forgot password like wireframe, see screen snap.

Screen Shot 2023-01-09 at 9 33 22 AM

straleyb:
Screen Shot 2023-01-09 at 12 02 45 PM

straleyb:
@azle I got it updated. Let me know if there is anything else that needs updated

azle:
@straleyb, looks great! Thank you!

CGillen (approved):

PR #2621, opened 2022-12-02
updates layout to match wireframe

straleyb:


app/views/catalog/_constraints.html.erb
```diff
@@ -22,7 +22,6 @@
 
 <div class="results-constraints-wrapper">
   <h2 class="sr-only top-content-title"><%= t('blacklight.search.search_results_header') %></h2>
-  <div class="result-count mw-lg-b col-md-3 col-sm-4"> <strong>Showing <%= @response.total_count %> works </strong></div>
+  <div class="result-count mw-lg-b col-md-3 col-sm-4"> <strong>Showing <%= ((@response.current_page * @response.limit_value) - (@response.limit_value)) + 1 %> to <%= (@response.current_page * @response.limit_value) %> of <%= @response.total_count %> work(s) </strong></div>
```
straleyb, line 25: I know this is some gross raw code but it calculates the current number to the last number of items currently displayed. I.E 8 to 16 of 1337. Do we mind this here or do we want it in an object? Seems fine for now here since its only used here, but im cool with whatever.

CGillen (approved):

PR #2638, opened 2022-12-12
updates search grid

straleyb:
There are a couple of questions lIke if we want to disable the header search bar placeholder text on the grid view. This would be quite difficult since the header is app wide and would need to know context based on which controller it routed through. It would then make it inconsistent app wide.

Second was if we wanted to hide the checkbox element and when it is clicked how we want it to operate. Do we want it to cover up the image and turn white as a border and grey out when checked or something else?

Third is how we want the search terms breadcrumb to show up. Right now they show up as whatever was put in the search box verbatim. do we want to force all text to lower case no matter what or keep it consistent with whatever someone searches.

@jsimic After these questions are figured out we should be good to go on this.


jsimic:
@straleyb 1) No, we do not want to hide the header search bar text. 2) No, we don't want to hide the checkbox, or change it or the border in any way once clicked. 3) No, we don't want to force lowercase. As entered is fine.

#1 and #2 are accessibility issues, and I consulted with Azle.

CGillen (approved):

PR #2669, opened 2023-01-03
Update jmespath to 1.6.2

CGillen:
Replaces #2636


CGillen:
Nvm #2636 is actually updatable

PR #2707, opened 2023-01-24
updates to display available collections to user and displays visibility

straleyb:
Screen Shot 2023-01-24 at 2 27 16 PM

This now displays the collections only a user can see. I also added the visibility of the work so the user can know which ones are visible only to them vs what everyone else can see. Works that are private cannot be seen by someone who isnt logged in.

Screen Shot 2023-01-24 at 2 28 58 PM

We can split this code out into its own view but hooking into the attribute_to_html was so indirect, especially when trying to extract visibilities so the users could know which ones were private vs public.


app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -12,7 +12,17 @@
...
-<%= presenter.attribute_to_html(:non_user_collections, render_as: :collection_link, html_dl: true, label: t("simple_form.labels.defaults.set")) %>
+<% docs = [] %>
+<% presenter.non_user_collections.each do |collection| %>
+  <% doc = SolrDocument.find(collection) %>
+  <% docs << doc if current_ability.current_user.can?(:read, doc)%>
+<% end %>
+<dt>Collections</dt>
+<dd><ul class="tabular">
+  <% docs.each do |doc| %>
+    <li class="attribute attribute-collections_label"><%= link_to doc.title_or_label, doc %><%= " (#{doc.visibility})" %></li><br>
```
CGillen, line 23: I'm not too excited about including the visibility, but maybe ping azle/julia with the images

CGillen (changes requested):

CGillen (approved):

PR #2801, opened 2023-02-22
Add support for ONS TFDDBasin on location CV

### LINKED ISSUE #2794: Add TFDDBasins ONS vocab to Location CV
### Descriptive summary

The TFDDbasins vocab from opaquenamespace.org needs to be added to the Location CV, particularly for use in the Water Basin field. This was a relatively late addition to the OD1 codebase and not originally on the MAP, but it's added now.

The freshwater treaties documents frequently use this vocab along with GeoNames in the Water Basin property.

Since the Location CV overrides the Hyrax one, I don't think it formally restricts what URIs can be added (beyond the GeoNames in type-ahead). So the opaquenamespace.org URIs can be added, but they currently fail at label parsing for in the Location CV:

```
Traceback (most recent call last):
       12: from (irb):316
       11: from (irb):316:in `each'
       10: from (irb):320:in `block in irb_binding'
        9: from app/indexers/document_indexer.rb:6:in `generate_solr_document'
        8: from app/indexers/generic_indexer.rb:17:in `generate_solr_document'
        7: from app/indexers/concerns/oregon_digital/indexing_dates_behavior.rb:9:in `generate_solr_document'
        6: from app/indexers/oregon_digital/deep_indexing_service.rb:21:in `add_assertions'
        5: from app/indexers/oregon_digital/deep_indexing_service.rb:43:in `fetch_external'
        4: from app/indexers/oregon_digital/deep_indexing_service.rb:43:in `each'
        3: from app/indexers/oregon_digital/deep_indexing_service.rb:44:in `block in fetch_external'
        2: from app/indexers/oregon_digital/deep_indexing_service.rb:52:in `block (2 levels) in fetch_external'
        1: from lib/hyrax/controlled_vocabularies/location.rb:44:in `rdf_label'
NoMethodError (undefined method `split' for #<RDF::Literal:0x1ea0c8("Colorado"@en)>)
```

OD1 example item link: https://od1.oregondigital.org/catalog/oregondigital:gb19fr166#page/1/mode/1up
OD2 work link that currently errors: https://prod.oregondigital.org/concern/documents/gb19fr166
example URI: http://opaquenamespace.org/ns/TFDDbasins/CLDO

### Expected behavior

TFDD basin URIs are allowed in fields powered by the Location CV, and labels fetched appropriately (no hierarchy or additions needed beyond what is fetched).

CGillen:
Fixes #2794


wickr (changes requested):
Looks like there's a few instances of 'TDFF' instead of 'TFDD'

lib/hyrax/controlled_vocabularies/location.rb
```diff
@@ -56,6 +59,27 @@ def rdf_label
...
+      def rdf_ons_label
+        labels = Array.wrap(self.class.rdf_label)
+        labels += default_labels
+        # OVERRIDE from rdf_triples to select only and all english labels
+        values = []
+        labels.each do |label|
+          values += get_values(label).to_a
+        end
+        eng_values = values.select { |val| val.language.in? %i[en en-us] if val.is_a?(RDF::Literal) }
+        return eng_values unless eng_values.blank?
```
straleyb, line 76: I think with these guard clauses there should be a space in between them.

lib/oregon_digital/controlled_vocabularies/extended_location.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module OregonDigital
+  module ControlledVocabularies
+    # Ethnographic Term object for storing labels and uris
+    class ExtendedLocation < Resource
+      def self.all_endpoints
+        [
+          OregonDigital::ControlledVocabularies::Vocabularies::OnsTDFFBasin
```
straleyb, line 9: Is this supposed to be TDFF or TFDD. The file controlled_vocabularies/vocabulaties/ons_tdffbasin.rb has it as tdff but the uri used is tfdd. These should be synced up properly id assume.

lib/hyrax/controlled_vocabularies/location.rb
```diff
@@ -56,6 +59,27 @@ def rdf_label
...
+      # But sometimes the first preferred label isn't in english, or all aren't
+      # Disable rubocop because just barely fails abcsize
+      # rubocop:disable Metrics/AbcSize
+      def rdf_ons_label
+        labels = Array.wrap(self.class.rdf_label)
+        labels += default_labels
+        # OVERRIDE from rdf_triples to select only and all english labels
+        values = []
+        labels.each do |label|
+          values += get_values(label).to_a
```
straleyb, line 73: I guess += does the same thing as << for arrays? I can see you are making a multidimensional array here but I would have expected to see << for pushing things into an array instead of +=

CGillen: I tried this and failed test. I think `#get_values` is returning an array, so array addition makes sense here

PR #2882, opened 2023-07-13
updates to sort file sets by alphabetical order.

straleyb:
This sorts the filesets. The filesets are sorted by the name of the file ingested. This is NOT sorting based on the name of the children ingested but THEIR filesets as well.


straleyb:
@CGillen This should be good to go now

CGillen (approved):

Issue #2804, opened 2023-02-23
Add 'Local Collection Name' to facets available on search results pages


petersec:
### Descriptive summary

SCARC staff have requested that the Local Collection Name facet be made available on search results pages. The two primary OSU history-focused collections in Oregon Digital - [Historical Images of Oregon State University](https://oregondigital.org/collections/osu-historical-images?locale=en) and [Historical Publications of Oregon State University](https://oregondigital.org/collections/osu-historical-publications?locale=en) - are both comprised of materials digitized out of many of SCARC's archival collections. As a component of daily work, it is important for SCARC staff to have the ability to narrow down search results within these two OD collections to aggregate content digitized out of specific local collections. Presently, the only spot where Local Collection Name is available is on the landing page for the two OSU history-focused collections noted above. Adding this functionality will also provide feature parity with OD1.

### Expected behavior

After conducting a search, I expect to see Local Collection Name as a faceting option available to me on the search results page. Upon selecting this facet, a roster of local collection names will appear as links. Selecting one of these links will allow me to aggregate all content digitized out of a specific local collection that was relevant to my original search and pulled into my search results.

lamtu1:
@CGillen From what I understand from this is the ticket is asking us rather than having the `Local Collection Name` (that we have listed in the above statement) on the homepage. We wanted to included in the search filter option facet like shown in the picture below correct?

This mean that we wanted to have a section called `Local Collection` that has the two links to the [Historical Images of Oregon State University](https://oregondigital.org/collections/osu-historical-images?locale=en) and [Historical Publications of Oregon State University](https://oregondigital.org/collections/osu-historical-publications?locale=en). When click on it, the search will filter out those two contents and find if the file have any matches with the keyword search.

<img width="1164" alt="Screenshot 2023-07-12 at 1 09 07 PM" src="https://github.com/OregonDigital/OD2/assets/111306182/1a3ec845-940a-49c4-8b39-701775a682d3">

petersec:
@lamtu1 @CGillen SCARC is requesting that Local Collection Name be added as a facet for all search results in the same manner as Copyright, Format, Media, etc. In other words, for all searches within OD2, Local Collection Name would be a facet available to users no matter their search terms or the collections that the search is hitting. SCARC will frequently use this for faceting within the Historical Images of OSU and Historical Publications of OSU set, but the facet should not be tied to those two sets in any specific way. Instead, it should be deployed for all content in Oregon Digital.

[...]

PR #2883, opened 2023-07-13
Add in facet with local collection

lamtu1:
Add in the facet of `local collection name` where display local collection we have as a search filter option to narrow down more items.


lamtu1:
Fixed for `issue:`   #2804

CGillen (approved):

Issue #2101, opened 2021-11-09
Create About Pages


jsimic:
### Descriptive summary

A series of pages containing information about Oregon Digital 2 need to be created and made public

### Expected behavior

Clicking on links contained in the site footer or on the home page should send the user to one of the informational pages. On any given page, a sidebar will ease navigation to other informational pages. There are a series of [wireframes](https://drive.google.com/drive/folders/1Yb1fUvcP2ap5TNo7F006ZiamXixwEJWb?usp=sharing) detailing the style and layout of each page:

- [x] Mission and Vision ([wireframe](https://drive.google.com/drive/u/2/folders/1mqIWCppDRLbrIiBd6lWr1SLVvM3qAyVC), [content](https://docs.google.com/document/d/1lixkmKiHSgqvmM8zfLYF_WLAOXpM-Zb516YYkRPpIJA/edit#))
- [x] About ([wireframe](https://drive.google.com/drive/folders/13qng4UmQLgPBgqPM25G0BCRTerLmqj_J?usp=sharing), [content](https://docs.google.com/document/d/1xcNuUjDyAarMaps5FuXXwoWVPS3DGszsYhtmpCEEMHs/edit))
- [ ] How to Use This Website  ([wireframe](https://drive.google.com/drive/u/2/folders/1jNoGNoMWagolnaZFq6jzRrFHeEUUVim9), content TBD)
- [x] Privacy Policy ([wireframe](https://drive.google.com/drive/u/2/folders/1E3w4RiYD8-jHeKYltSsqIDSGoARP6RXo), [content](https://docs.google.com/document/d/1NiBt64ZlTxeiC94xpCys-_XwqZGRRZ6YTpweCnPx4w8/edit))
- [x] Terms of Use ([wireframe](https://drive.google.com/drive/folders/1Q55-xsF2Mfwm2zf8LQR51X1Z7XqcYSDb?usp=sharing), [content](https://docs.google.com/document/d/1CGHaZ3mDRMPorGDZfJSp9sv_JWliEZ9lgoTW8Iob4AI/edit))
- [x] Copyright and Usage Requests ([wireframe](https://drive.google.com/file/d/1c6WCbYUocU4wh8ujC_xFjkjmSe0LrPP6/view?usp=sharing), ([content](https://docs.google.com/document/d/1HC_YBgFUIgv5TPUcLrSTL9N4kLEs9tjRAMBneChhSK8/edit))
[...]

PR #2893, opened 2023-07-20
Add in new content to the how-to page

lamtu1:
`In Progress:` Add in content on the `how-to` page with new content and layout.


lamtu1:
`Fixed for Issue:`   #2101

lamtu1:
`Quick Video:` Demoing the How-To Use page
https://github.com/OregonDigital/OD2/assets/111306182/1cbe603b-85ce-4812-8966-9b64433f4387

straleyb:
@lamtu1 Is this ready to go or still WIP?

lamtu1:
@straleyb I think it is ready to go. We can have this up for now, and if anyone have anymore suggestion or edit, I can alter it later on.

app/views/oregon_digital/about/use.html.erb
```diff
@@ -1,13 +1,76 @@
+<html id="use_page">
+<body>
 <h2>How to Use Oregon Digital</h2>
 
-<h3>Searching and Faceting</h3>
+<br />
+
+<%# LIST: Showing Table of Content for navigation %>
+<h3>Table of Contents</h3>
+<ol class="use_toc">
```
CGillen, line 9: Good use of `ol`

CGillen (changes requested):
What was your reasoning for styling this page different than the rest of the about pages?

lamtu1:
> What was your reasoning for styling this page different than the rest of the about pages?

I did this styling different as I was following along the google docs requirement on the ticket. Here is the link to the [Google Doc](https://docs.google.com/document/d/17xaOFMCH-_5s-txlTIzroFF3369Q-chAm-EGau5Oay8/edit#heading=h.byqkkjiuntp6).

I was not sure what the stylization is for the about page, so I just layout based on this requirement as follow in the link above.

CGillen:
The layout looks great, but I don't see where they mentioned they needed different CSS-type styling.
You should also almost never use `display: table` It usually means you're either working with a hack, or have messed up your `<table>` somehow. Either way, it's really bad for a11y

lamtu1:
@CGillen Ah I see, I was following a tutorial on how to use CSS to display the number like in the demo video on the table of content. And I created the CSS page to isolate the content only on that page out like some spacing issue between `<p>` and `<h3> or <h2>` tag and make it only affecting on that `how-to` page. But I can rework it on the `display: table`and if we are sticking with no CSS page, I can rework it as well to use our CSS only.

CGillen:
If you need a separate css file just for the how-to page, that's fine. Just try to be aware of the font/text styling, to keep it consistent with the rest of the site. There should be some helper classes for some common resizing/styling fonts here
https://github.com/OregonDigital/OD2/blob/master/app/assets/stylesheets/oregon_digital/fonts.scss

stuff like `.lato-p`, or `.mw-lg`. Those aren't ones I specifically want you to use somewhere, they're just examples of the kinds of classes that we commonly use to keep fonts consistent

lamtu1:
I will do that. I will fix that `display: table` in the css and beware of the other pages, and look how to apple consistency based on the `fonts.scss` file. Thank you for the feedback @CGillen

CGillen:
Can you post another screenshot? I'm worried some text is still serif and that just doesn't feel right. All font should be `lato`

lamtu1:
Here some screenshot

Screenshot 2023-08-02 at 2 39 30 PM

Screenshot 2023-08-02 at 2 39 45 PM

CGillen (approved):

PR #2897, opened 2023-07-27
Bulkrax patch from v5.1

### LINKED ISSUE #2896: bulkrax exporter can't use importer as source
### Descriptive summary
Created an importer to import 9 images, then tried to export them using the importer as the source. This process failed two attempts, the first time it remained stuck in pending and reported not finding any entries. The second time it reached status failed, and reported an error in the code.

### Expected behavior
User should be able to export based on a prior importer.

### Related work

### Accessibility Concerns

lsat12357:
fixes #2896


CGillen (approved):
Out of curiosity, and I'm probably just out of the loop, why can't we update bulkrax?

PR #2900, opened 2023-07-31
Use the same idea from #work_presenters for #file_set_presenters

### LINKED ISSUE #2863: Reordering of multiple filesets is not reflected in viewer
### Descriptive summary

_Background: The Metadata Team discussed and agreed that, in certain cases, it will be preferable to create single works with multiple filesets rather than true complex objects._

Multiple filesets attached to a work can be managed using the File Manager option in the UI: setting which is the Thumbnail and the Representative Media for the work, and applying alphabetical sorting. However, changes to order made through the File Manager are not mirrored in the UV for the work.  

In this example on Staging -- https://staging.oregondigital.org/concern/images/q524jn80f?locale=en -- the black and white image was added first and the color image added second. After initially saving, I used the File Manager to edit the files: I set the color image as the thumbnail, and renamed the files from "_front" and "_back" to "a" and "b" so that I could sort them alphabetically. After that edit, the order is shown correctly at the bottom of the work, and the desired thumbnail appears in search results, etc, but the black and white image remains first in the viewer. I cannot modify this order elsewhere in the UI.

image

### Expected behavior

Multiple files attached to a single work can be reordered using the File Manager, and those changes will apply to the order that files are displayed in the viewer.

Having additional sorting controls in the File Manager, beyond initial upload order and alphabetical, would be an enhancement.

### Related work

### Accessibility Concerns

CGillen:
Fixes #2863

### DIFF
--- a/app/presenters/oregon_digital/iiif_presenter.rb
+++ b/app/presenters/oregon_digital/iiif_presenter.rb
@@ -18,19 +18,27 @@ module OregonDigital
     end
 
     # rubocop:disable Metrics/AbcSize
+    # TODO: DRY this up with #work_presenters?
     def file_set_presenters
-      presenters = []
-      file_sets.each do |fs|
+      file_set_ids = (ordered_ids & file_sets.map(&:id))
+      return [] if file_set_ids.empty?
+
+      solr_query = Hyrax::SolrQueryBuilderService.construct_query_for_ids(file_set_ids)
+      solr_response = Hyrax::SolrService.get(solr_query, rows: 10_000)
+      document_hash = array_to_hash_response(solr_response)
+
+      presenters = file_set_ids.map do |id|
+        fs = SolrDocument.new(document_hash[id])
         deriv_type, presenter_class = file_set_presenter_info fs
 
         urls = file_set_derivatives_service(fs).sorted_derivative_urls(deriv_type)
 
-        urls.each_with_index do |derivative, i|
+        urls.each_with_index.map do |derivative, i|
           label = urls.length > 1 ? page_label(fs.parents.first.title_or_label, i) : fs.parents.first.title_or_label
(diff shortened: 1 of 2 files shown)


CGillen:
@straleyb Can you review this? It should just be a copy+paste for the `#work_presenters` with some minor changes

app/presenters/oregon_digital/iiif_presenter.rb
```diff
@@ -18,21 +18,31 @@ def manifest_metadata
     end
 
     # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength
+    # TODO: DRY this up with #work_presenters?
     def file_set_presenters
-      presenters = []
-      file_sets.each do |fs|
+      file_set_ids = (ordered_ids & file_sets.map(&:id))
```
straleyb, line 24: I wonder if we need the ordered_ids here. Since its just filesets, we should be able to get away with just those. Is there a reason why we would need to grab just filesets that are included in ordered_ids as well?

CGillen: I didn't check to see if `file_set_ids` order actually updates when they're re-arranged. If they are then we should be able to get away with not using `ordered_ids`, otherwise I think it's important

CGillen:
```
>> file_sets.map(&:id)
=> ["f0n009w228g", "f2n009w228g", "f4n009w228g", "f6n009w228g", "f8n009w228g", "f10n009w228g"]

>> ordered_ids & file_sets.map(&:id)
=> ["f8n009w228g", "f4n009w228g", "f0n009w228g", "f2n009w228g", "f6n009w228g", "f10n009w228g"]

```

`ordered_ids & file_sets.map(&:id)` is the correct order

straleyb (approved):

Issue #2879, opened 2023-07-13
Change Character Length of Title For Child Works in Viewer


KevinJonesMeta:
### Descriptive summary

Change length of displayed title on child works from 7 characters to 50 characters in contents list in viewer to allow user to see full titles of child works when viewing parent work of compound objects.

### Expected behavior

50 characters of title of a child work display in contents list in viewer for compound objects.

### Related work

https://github.com/OregonDigital/OD2/issues/2880

lamtu1:
@KevinJonesMeta would you like the Child Works in Viewer to display the entire title or would you have set limit if a title is too long, then having a `50 characters limit` and hide the rest of the title where you have to hover over it to see the rest of the name?

KevinJonesMeta:
@lamtu1 having a 50 character limit and hiding the rest is great. if we discover with this change that we would like to increase the character limit will that be significant work to change? We don't want them to be really long but we do want users to see enough of the title on child works to tell which is which.

KevinJonesMeta:
QA pass. Expected character count is displayed on child works on staging.

PR #2901, opened 2023-07-31
Change display of UV to display more characters

lamtu1:
`In Progress:` Change the character display to show all the `label` content rather than only displaying 7 characters.


lamtu1:
`Fix for Issue:` #2879

CGillen (approved):

PR #2907, opened 2023-08-03
cleaning up method calls

### LINKED ISSUE #2902: bulkrax not reporting child works in export
### Descriptive summary
Bulkrax is finding the filesets, but in the case of a cpd, not putting any child works in the children column.

### Expected behavior
pids of child works should be listed with separators in the children column of an export for a given cpd work

### Related work

### Accessibility Concerns

lsat12357:
fixes #2902

### DIFF
--- a/config/initializers/bulkrax.rb
+++ b/config/initializers/bulkrax.rb
@@ -120,6 +120,29 @@ Bulkrax::CsvEntry.class_eval do
     end
   end
 
+  # changing call from member_work_ids to member_ids as the former does not return results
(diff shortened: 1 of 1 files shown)


lsat12357:
The problem is that bulkrax is using member_work_ids to get child works, but member_work_ids requires a solr field that OD does not use.  As work.member_work_ids is not commonly used anywhere, but member_ids is, the straightforward approach seems to be to swap them. At the moment, order is only relevant in the children column when importing new parents, so member_ids is sufficient.

CGillen (approved):

PR #2905, opened 2023-08-02
Search performance enhancements

CGillen:
Batch solr look ups
Improved SolrDocument fileset/child lookup

PR #2909, opened 2023-08-04
Implement SolrDoc array lookup

CGillen:
Following #2905


CGillen:
@straleyb This replaces all the loop + solr queries I could find

straleyb (approved):

Issue #1719, opened 2021-06-24
Add 'Edit' button to Collection show page


wickr:
### Descriptive summary

A Collection show page needs an 'Edit' button, for those with access, similar to how Works have one. Currently you have to go through the Dashboard.

### Expected behavior

Collection show page has an 'Edit' button, probably to the right of 'View All' link.

lamtu1:
@wickr I have a PR up for this and wanting to clarify. Do we only allow people who are admin access the `Edit` collection page or people who created their own collection can have that button `edit` show up too?

wickr:
@lamtu1 yes people who created their own collection should have this too, along with admins. 

This isn't changing the access levels at all, just making a button available if they do have access.

CGillen:
QA: Deployable.
Adding an edit button for User Collections is slightly more difficult since it requires a different functionality on the button. Users have access to the My Collections interface which has the edit button, I vote we leave it at that for non-privileged users.

shieldsb:
QA pass.

PR #2920, opened 2023-08-14
Add in an edit button to Collection page

lamtu1:
Add in a button to go to the `edit` section for Collection rather than going to the `Dashboard`.

### DIFF
--- a/app/views/hyrax/collections/show.html.erb
+++ b/app/views/hyrax/collections/show.html.erb
@@ -90,7 +90,7 @@
   <%= render 'shared/colored_divider', color: 'lime-green' do %>
     <a href="<%= view_all_url %>" class='btn btn-lg btn-primary collection-landing-button col-sm-10'>View All</a>
     <%# NEW BUTTON: Add in an edit button to collection rather than going through the dashboard %>
-    <% if user_signed_in? && current_user.admin? %>
+    <% if user_signed_in? && (current_user.admin? || current_ability.can?(:edit, @presenter.solr_document)) %>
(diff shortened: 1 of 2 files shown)


lamtu1:
fixes #1719

lamtu1:
`SCREENSHOT:` Showing the `edit` button in Collection

Screenshot 2023-08-14 at 4 47 08 PM

app/views/hyrax/collections/show.html.erb
```diff
@@ -90,7 +90,7 @@
   <%= render 'shared/colored_divider', color: 'lime-green' do %>
     <a href="<%= view_all_url %>" class='btn btn-lg btn-primary collection-landing-button col-sm-10'>View All</a>
     <%# NEW BUTTON: Add in an edit button to collection rather than going through the dashboard %>
-    <% if user_signed_in? && current_user.admin? %>
+    <% if user_signed_in? && (current_user.admin? || current_ability.can?(:edit, @presenter.solr_document)) %>
```
CGillen, line 93: Patron users can also create collections & don't have access to the dashboard to edit their collections. You need to make sure this isn't a User Collection type collection. We usually do that with something like:
`collection_type.machine_id == 'user_collection'` You'll have to edit that to properly get the collection_type

CGillen (changes requested):

app/views/hyrax/collections/show.html.erb
lamtu1: @CGillen How does the `user_collection` see on their end when they edit? Is it just this?

Screenshot 2023-08-18 at 10 30 21 AM

CGillen (approved):

PR #2933, opened 2023-08-25
AZ Topic facet filter

### LINKED ISSUE #1133: Improved more facets A-to-Z view
### Descriptive summary
In the 'more' pop-over in the facet, one of the sorting options is A-to-Z.
The display should include an alphabet select at the top to further filter the results by first letter.

### Expected behavior
A-to-Z has an easy way to jump to facet values based on first letter.

### Related work
N/A

### Accessibility Concerns
Alphabet list are notoriously bad for a11y. They're a pain to skip pass and provide little context for what's going on as you're tabbing through. Make sure there is at least a header immediately after the list so keyboard users can jump pass it by headers.

Also, live updating content is difficult to get right. Be sure to follow [live-regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) and do some research on properly implemented a-to-z lists.

megwalt:
Rough working implementation of an A-Z index filter (closes #1133 ) in the Filters > Topic > More modal.

Default "All Topics":
Screenshot 2023-08-23 at 4 30 09 PM

Topics filtered by letter:
Screenshot 2023-08-23 at 4 30 32 PM

Topics filtered by letter (case: no topics):
Screenshot 2023-08-23 at 4 30 38 PM

Accessibility work that’s happened so far:
- A-Z Links are each list items so that screen reader users know how long the list is + make an informed decision.
- Wrapped a nav tag around the list with an aria-label for easier navigation and clarity of purpose.
- Added a heading after the navigation for a screen reader user to skip to. 
- Designated the heading as an aria-live region that screen readers will read after a live update. Only the heading is read to prevent an uninterrupted, not-useful reading of every topic on screen. 

Needs feedback/more work:
- Content of live heading. Currently "Topics starting with "{letter}"", but that doesn’t seem in line with the rest of the site. The description of what has changed on update is essential so updating the header to be more concise would mean this description would go somewhere else (visible or not). 
- After clicking on one of the A-Z links, the focus of the screen reader returns to the modal (as opposed to staying in the nav). Is this expected behavior? The same is true when clicking the "next" and "prev" links meaning the user has to skip past the nav to get to the list of topics. The heading after the nav should help with this.
- I need to implement something different for the case where no topics starting with a certain letter exist? 

In the future:
- Would a text input field be a viable option as a filter instead of the A-Z index? Following something inspired by the proposal in https://www.scottohara.me/blog/2022/02/05/dynamic-results.html. An implemented example of this found here: https://a11ysupport.io/.

### DIFF
--- a/app/views/catalog/_facet_index_navigation.html.erb
+++ b/app/views/catalog/_facet_index_navigation.html.erb
@@ -1,4 +1,4 @@
-<nav id="az-index" aria-label="'A' to 'Z' topic index">
+<nav id="az-index" aria-label="'A' to 'Z' topic filter">
   <ol class="pagination pagination-sm justify-content-center">
     <li class="page-item <%= 'active' if @pagination.prefix.blank? %>" >
       <a class="page-link" data-href="<%= url_for(@pagination.params_for_resort_url('index', @search_state.to_h.except(@pagination.request_keys[:prefix]))) %>" data-ajax-modal="preserve" aria-controls="facet_list_header">
--- a/app/views/catalog/_facet_index_navigation.html.erb
+++ b/app/views/catalog/_facet_index_navigation.html.erb
@@ -1,0 +1,16 @@
+<nav id="az-index" aria-label="'A' to 'Z' topic index">
+  <ol class="pagination pagination-sm justify-content-center">
+    <li class="page-item <%= 'active' if @pagination.prefix.blank? %>" >
+      <a class="page-link" data-href="<%= url_for(@pagination.params_for_resort_url('index', @search_state.to_h.except(@pagination.request_keys[:prefix]))) %>" data-ajax-modal="preserve" aria-controls="facet_list_header">
+        <%= t('blacklight.search.facets.all') %>
+      </a>
+    </li>
+    <% @facet.index_range.each do |letter| %>
+      <li class="page-item  <%= 'active' if @pagination.prefix == letter %>">
+        <a class="page-link" data-href="<%= url_for(@pagination.params_for_resort_url('index', @search_state.to_h.merge(@pagination.request_keys[:prefix] => letter))) %>" data-ajax-modal="preserve" aria-controls="facet_sorted_list">
+          <%= letter  %>
+        </a>
+      </li>
+    <% end %>
+  </ol>
+</nav>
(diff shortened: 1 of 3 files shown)


app/views/catalog/facet.html.erb
```diff
@@ -12,13 +12,21 @@
 
 </div>
 <div class="modal-body">
+  <div role="region" id="facet_list_header" aria-live="polite">
+    <% if @pagination.prefix && @display_facet.index? %>
+      <h4>Topics starting with "<%= @pagination.prefix %>" </h4>
```
CGillen, line 17: Change `Topics` to something that holds the facet title, `@display_facet` maybe?

app/views/catalog/_facet_index_navigation.html.erb
```diff
@@ -0,0 +1,16 @@
+<nav id="az-index" aria-label="'A' to 'Z' topic filter">
```
CGillen, line 1: Change `Topics` to something that holds the facet title, `@facet` maybe?

app/views/catalog/facet.html.erb
```diff
@@ -12,13 +12,21 @@
...
+    <% elsif @display_facet.index? %>
+      <h4><%= t('blacklight.search.facets.all') %> Topics</h4>
```
CGillen, line 19: Here too

CGillen (changes requested):
Accessibility work that’s happened so far:
- All this is GREAT! Thank you!

Needs feedback/more work:
1. I think that's in line. However, I'd like it more dynamic. Right now it only works for `Topic` but we might want to use this for other facets in the future, especially if we get a type-ahead
2. I hate this too, but it's the current functionality so we're not losing anything at least 😢 
3. This is probably fine

In the future:
- Yes! Please! Can you open an enhancement ticket?

app/views/catalog/facet.html.erb
```diff
@@ -14,9 +14,9 @@
 <div class="modal-body">
   <div role="region" id="facet_list_header" aria-live="polite">
     <% if @pagination.prefix && @display_facet.index? %>
-      <h4>Topics starting with "<%= @pagination.prefix %>" </h4>
+      <h4><%= facet_field_label(@facet.key) %>s starting with "<%= @pagination.prefix %>" </h4>
```
CGillen, line 17: Strings in Ruby actually have a `#pluralize` method. Its WILD!
```
'post'.pluralize             # => "posts"
'octopus'.pluralize          # => "octopi"
'sheep'.pluralize            # => "sheep"
```

CGillen (changes requested):

app/views/catalog/facet.html.erb
megwalt: That's awesome, fixed!

CGillen (approved):

PR #2954, opened 2023-09-29
Valkyrize round 1

CGillen:


CGillen:
@straleyb 
The big main changes are:

`ActiveFedora::Base#find` to `Hyrax.query_service#find_by_alternate_identifier(alternate_identifier:)`. This is the main change to do a valkyrie look up

`item.to_sipity_entity` to `Sipity::Entity(item)` This is the new way to lookup Sipity Entities apparently, since Valkyrie objects don't have `#to_sipity_entity`

`work.file_sets` becomes `work.member_ids.map(&:id)` with a `Hyrax.query_service` lookup for each id. A `Valkyrie::ID` object doesn't easily resolve to the id string. You have to either call `#id` or `#to_s`. So you end up with a lot of `valkyrie_resource.id.id` 🙄 

`work.update_index` to `Hyrax.index_adapter.save(resource: work)`. It's just the new way to index something

I think that's mostly it

straleyb (approved):

PR #2955, opened 2023-10-02
Prevent user collections and use correct collection sort on homepage

CGillen:
@wickr For the home page explore collection issues this morning


wickr (approved):

CGillen:
@shieldsb QA for this ticket is making sure that only digital collections are visible on the homepage for all users. No user collections and no OAI sets

shieldsb:
Created a new user collection and it did not display in the featured collections section

Issue #2941, opened 2023-09-01
Change Bulkrax Importer and Exporter list views default to sort by most recent


wickr:
### Descriptive summary

Now that we have more than just a few Importers and Exporters, alpha sort of the lists doesn't make much sense. The most recent ones should appear at the top. Each column still has the 'sort by' icons to choose later, this would only affect the default.

### Expected behavior

Importer view default to sort by Last Run
Exporter view default to sort by Date Exported

### Related work

Link to related tickets or prior related work here.

### Accessibility Concerns

Add any information here to indicate any known or suspected accessibility issues for this ticket

KevinJonesMeta:
QA pass. Tested on Chrome and Edge.

Importer view defaults to sort by Last Run
Exporter view defaults to sort by Date Exported

PR #2957, opened 2023-10-04
Add in default sorting to export & import

lamtu1:
Overriding the original layout to add in default sorting for both the `importer` & `exporter` for Bulkrax.

### DIFF
--- a/app/views/bulkrax/exporters/index.html.erb
+++ b/app/views/bulkrax/exporters/index.html.erb
@@ -15,7 +15,7 @@
 <div class="panel panel-default">
   <div class="panel-body">
     <% if @exporters.present? %>
-      <div class="table-responsive">
+      <div class="table-responsive" id="export-table">
         <table class="table table-striped datatable">
           <thead>
             <tr>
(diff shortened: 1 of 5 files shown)


lamtu1:
fixes #2941

lamtu1:
`DEMO:` Showing the sorting default in bulkrax.

https://github.com/OregonDigital/OD2/assets/111306182/c4116777-23d9-442b-ad92-427be62a4672

app/views/bulkrax/exporters/index.html.erb
```diff
@@ -0,0 +1,76 @@
...
+          </tbody>
+        </table>
+      </div>
+      <% else %>
+        <p>No exporters have been created.</p>
+      <% end %>
+  </div>
+</div>
+
+<script>
```
straleyb, line 59: Can this script be moved over into a js file? That would be more close to what we have normalized in the application.

app/views/bulkrax/importers/index.html.erb
```diff
@@ -0,0 +1,83 @@
...
+        </table>
+      </div>
+      <% else %>
+        <p>No importers have been created.</p>
+      <% end %>
+  </div>
+</div>
+
+<script>
+  $(function() {
```
straleyb, line 67: Same here

app/views/bulkrax/exporters/index.html.erb
lamtu1: Will do! since both of the file are using the same order, having one script js file should be good.

straleyb (approved):

Issue #2937, opened 2023-08-31
Add Download button for uploaded CSV or zip of bulkrax Importer


wickr:
### Descriptive summary

Bulkrax importers show the file path of the uploaded CSV or zip, but no easy way to then download the file. It would be helpful to have a Download button next to the file path. 
This would be useful for debugging and solving metadata issues, and be more direct than showing a Google Sheet or Excel file since it's exactly what the server is trying to parse. 

![image](https://github.com/OregonDigital/OD2/assets/2293544/a8aa6ffd-907b-4d57-9366-e1eea26e03b8)

### Expected behavior

Download button in Importers for original, uploaded files.

### Related work

Link to related tickets or prior related work here.

### Accessibility Concerns

Add any information here to indicate any known or suspected accessibility issues for this ticket

KevinJonesMeta:
QA Pass. Download button appears in expanded Parser Fields box on staging.

![image](https://github.com/OregonDigital/OD2/assets/57114731/d5b601ef-dd16-4284-8f00-ce755e9b4e4e)

PR #2960, opened 2023-10-06
Add in download button for importer

lamtu1:
Add in a way to fetch the importer file and make easy access to get the file out of import

### DIFF
--- a/app/controllers/bulkrax/importers_controller.rb
+++ b/app/controllers/bulkrax/importers_controller.rb
@@ -3,6 +3,7 @@
 require_dependency 'bulkrax/application_controller'
 require_dependency 'oai'
 
+# OVERRIDE: Override controller to add in new method for download
 module Bulkrax
   # rubocop:disable Metrics/ClassLength
   class ImportersController < ApplicationController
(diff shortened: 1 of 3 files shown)


lamtu1:
fixes #2937

straleyb:
@lamtu1 This looks good. I wonder if we need to bring in that entire controller or if it would be better if we created a hack of some kind to just open the class using class eval and then added our method in there.

I think it would be that or creating a controller that extends that one and then adds our method in.

Either would be fine on my end. The first one is probably easier. @CGillen Any input on that idea?

lamtu1:
`DEMO:` Showing that files can be download

https://github.com/OregonDigital/OD2/assets/111306182/c8444fb6-d12e-4ff0-8417-15cdbf118bf3

lamtu1:
> @lamtu1 This looks good. I wonder if we need to bring in that entire controller or if it would be better if we created a hack of some kind to just open the class using class eval and then added our method in there.
> 
> I think it would be that or creating a controller that extends that one and then adds our method in.
> 
> Either would be fine on my end. The first one is probably easier. @CGillen Any input on that idea?

Corey and I have a discussion on this subject, and we decided that it would be easier to approach this way for the importer by adding in a function to download. We did try to explore the other method to call the controller in other areas and seems to be a weird doozy one to do, so we end up deciding on this way as the best option for right now.

straleyb (approved):

PR #2967, opened 2023-10-12
Add collection permissions creation after save

straleyb:
We need to grab a user here for the service to work. This should create the permissions for a freshly saved collection.


lsat12357:
I see you added Hyrax::Collections::PermissionsCreateService.add_access(collection_id: c.id ...) to the function prep_collection_type(). But the c.id in that function is not the id of a Collection, it's the id of a CollectionType. The body of the rake task is where the collections are made, so after line 20, you'd have a freshly saved collection.

straleyb:
Alright got it updated @lsat12357 Thanks for working with me on this!

lsat12357 (approved):

Issue #1517, opened 2021-02-08
Expose my collection download on collection landing page


CGillen:
### Descriptive summary
There should be a button (wireframe to come) for users to download User Collections which have been shared (public). This will allow professors and researchers to create collections and their students to download course materials or otherwise interesting collections of works.

### Expected behavior
A user can see a download button and use it on `User Collections` which they have download access to. Users should be able to download any collection which is public/shared. Works that the user shouldn't be able to view, shouldn't be added to the zipped download (standard behavior already)

shieldsb:
QA fail. Download button not available on collection landing page.

CGillen:
@shieldsb @jsimic It looks like @lamtu1's update was to add Digital Collections, but this ticket is specifically for User Collections. If we want to do Digital Collections we'll want to architect this differently since they tend to be huge

shieldsb:
No, it didn't appear on the user collections page either.

lamtu1:
I'm currently checking on staging and user collection download works for me

lamtu1:
If `digital collection` is not part of it. Then I can go in remove it and update the code to do `user collection`.

CGillen:
the ticket requires the user's collection to be shared (public with the green share link on my collections dashboard). This seems to be working for me too.
Does this ticket need to be reworked?

lamtu1:
I don't think so, I changes back to the user collection and seem to be working fine on staging. But I will fixed on some of the feedback you requested in the PR and make update to it and will let you know once it is ready

lamtu1:
Hi @jsimic I have a question about the user collection download. In the future, are we ever will have a time we want to allow full/high res for collection?

If full size is ever going to be allowed, we need exactly who and under what conditions they can download full size. And downloading full size works is already figured out, so if we dynamically decide to add full or low works based on that, we won't be able to re-architect it later to massively conserve resources in the future. Right now downloads for collections are slow, costly to the system, prone to crashing, and not resumable if things crash.

jsimic:
@lamtu1 The consensus from the metadata team is no, we do not want to allow hi-res downloading.

shieldsb:
@CGillen @lamtu1 When I try to download a collection from the collections landing page, I receive an error message. 
![image](https://github.com/OregonDigital/OD2/assets/53542788/0cb25ab2-1407-4c96-9b9c-1b8e7607d275)

shieldsb:
Download does not prompt with option for high res images

PR #2971, opened 2023-10-13
Add download to the user collection

lamtu1:
Add in a download button to all the works save in the `user collection` if it is make for the public and it is sharable.


lamtu1:
fixes #1517

straleyb (approved):

PR #2973, opened 2023-10-24
Hyrax 5.0.0.rc2 OD2

straleyb:


straleyb:
There is a problem here with libressl-dev. There are conflicts in our installed libraries with libressl-dev conflicting with openssl. Its all used for installing mariadb. There are some workarounds I saw which rely on installing openssl, building a library called cryptography, then building mariadb, then uninstalling openssl and then installing libressl. Its wonky and idk if thats what we need to do.

straleyb:
We have libressl building fine. Its just libressl-dev

PR #2981, opened 2023-11-16
Export query

### LINKED ISSUE #2938: add local collection as a query in bulkrax exporter
### Descriptive summary
OSU makes a lot of use of local collection, and enabling this feature will make it possible to have smaller exports.

### Expected behavior
User will be able to input a local collection as the source for an export. 

### Related work

Making this change will be simplified by waiting for the upgrade to 5.2.

### Accessibility Concerns

lsat12357:
fixes #2938


config/initializers/bulkrax.rb
```diff
@@ -83,7 +83,6 @@
     v['from'] = [k]
   end
   fieldhash_csv['bulkrax_identifier'] = { from: ['original_identifier'], source_identifier: true }
-  fieldhash['visibility'] = { from: ['visibility'] }
```
CGillen, line 86: Is this intentional or a local hack that slipped in?

CGillen (approved):

PR #2992, opened 2023-11-29
Fix export query

### LINKED ISSUE #2938: add local collection as a query in bulkrax exporter

lsat12357:
fixes #2938

fixes label and display source probs. Query syntax stopped working evidently over use of an asterisk in the params, but it's not needed, so just removing it.

### DIFF
--- a/app/parsers/bulkrax/parser_export_record_set.rb
+++ b/app/parsers/bulkrax/parser_export_record_set.rb
@@ -208,7 +208,7 @@ module Bulkrax
 
     class LocalCollection < Base
       def works_query
-        "local_collection_name_tesim:*#{local_coll_id} #{extra_filters}"
+        "local_collection_name_tesim:#{local_coll_id} #{extra_filters}"
       end
 
       def local_coll_id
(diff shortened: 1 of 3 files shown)


app/parsers/bulkrax/parser_export_record_set.rb
```diff
@@ -208,7 +208,7 @@ def collections_query
 
     class LocalCollection < Base
       def works_query
-        "local_collection_name_tesim:*#{local_coll_id} #{extra_filters}"
+        "local_collection_name_tesim:#{local_coll_id} #{extra_filters}"
```
wickr, line 211: the `_tesim` field might be too permissive and match on tokenized splits of the URI suffixes (collection IDs) so I think we should stick with `_sim` for an exact match

lsat12357:
Screenshot at 2023-11-29 12-27-37

Issue #2634, opened 2022-12-08
Webform mismatch between uri and label


jsimic:
### Descriptive summary

When editing a record using the webform, I noticed in the Subject list that the labels of fetched terms do not match the URIs displayed. This may cause some confusion when adding or deleting terms.

<img width="688" alt="subjectURImismatches" src="https://user-images.githubusercontent.com/6344086/206531461-0e4b5a3c-7bfe-4dd1-b076-ac05312eaa15.png">

### Expected behavior

Labels and URIs should be matched. There may be mismatches is other fields, I haven't checked.

### Related work

[#500](https://github.com/OregonDigital/OD2/issues/500), [#34](https://github.com/OregonDigital/OD2/issues/34)

jsimic:
@lamtu1  id.loc.gov is the Library of Congress. An overwhelming number of subject headings use it as do all the Creator fields and others.

lamtu1:
@jsimic I just double check on the `Library of Congress`, it looks like it seems good on my end when double check on the URI and the label. But I do see that the Library of Congress update some of the vocab and changes the location in some of the word that use in the ticket. Below the images show:

<img width="686" alt="Screenshot 2023-11-14 at 2 40 48 PM" src="https://github.com/OregonDigital/OD2/assets/111306182/26f89fbc-ecba-425d-b3b4-de2b03aef871">

I think some of the event why some `Subjects` doesn't match up might be due to `LOC` move vocab around to a different URI. But I think this ticket can be QA as I am not seeing thing out of the ordinary.

jsimic:
@lamtu1 I think it's unlikely that LOC moving terms around is the culprit. That usually happens if they move between /subjects/ and /names/ or a term is depreciated and a new record with a new ID is created to replace it. I'm hoping that this got taken care of earlier in development. The original record I was looking at still has the mismatch however: https://oregondigital.org/concern/images/df69h2447/edit?locale=en#Subjects. The first line with the label "Mythology, Classical" has the URI for "Gods" and vice versa.

jsimic:
@lamtu1  The line in the record above labeled architectural interiors has a URI for a depreciated LOC term. That might be what happened--the label for the LOC term didn't get fetched because the record was depreciated and it just assigned the label of the next term to the URI, leaving the last URI without a label. We need to figure out some way of searching for these so they can be corrected...

lamtu1:
@jsimic Would you want a separate ticket on this matter? and have it link to this ticket?

[...]

PR #3006, opened 2023-12-07
Add in special way to get correct uri & label

lamtu1:
`FEATURE:` Add in a special way to get the URI & label correctly when rendering form on edit.

`TASK:` Task list for implementing the `label$uri` in OD2

- [x] Create `TriplePoweredService` for OD2 for functionality fetching and put as `label$uri` and create the `spec` for it
- [x] Build `Triple Powered Properties for solr doc` function in Generic Indexer
- [ ] ~(Maybe?) Need to add `FormMetadataService`~
- [x] Add the method to `generate_solr_document` to save it to Solr
- [x] Create a `label parser` class and write the `spec` for it
- [x] Add in the label functionality to `solr_document` to access the function on the variable
- [ ] (Maybe?) Add to `catalog_controller` as a `search_field`


lamtu1:
fixes #2634

CGillen (changes requested):
Unfortunately, this is a wider systemic issue than just `subject`, so a special case wouldn't be the best here.
This solution will also kick off a network call for each controlled term, whether to our cache or the source that's too slow for a page load. We **_have_** to do this work on index/save

You might find a better solution in placing a placeholder value in the solr index where a label couldn't be fetched on index or something else at index time. Alternatively, we could discuss a transition plan for using split label indexing like SA@OSU which might be the ultimate goal

lamtu1:
@CGillen Right now, I can try to find a way to fix this temporary with a placeholder on index time, but from what you said, I think it is best to create a new ticket to overhaul the `uri & label` for the whole system wide like in SA@OSU which will be efficient and less broken-ness on the fetching end if they are depreciated link.

CGillen:
I agree, lets go with a place holder `Unable to fetch label` and I'll open a ticket for transitioning to `$` split indexing

lamtu1:
`TASK:` Task list for implementing the `label$uri` in OD2

- [ ] Create `TriplePoweredService` for OD2 for functionality fetching and put as `label$uri` and create the `spec` for it
- [ ] Build `Triple Powered Properties for solr doc` function in Generic Indexer
- [ ] (Maybe?) Need to add `FormMetadataService` if field have `Other` in `Triple Powered Properties for solr doc`
- [ ] Add the method to `generate_solr_document` to save it to Solr
- [ ] Create a `label parser` class and write the `spec` for it
- [ ] Add in the label functionality to `solr_document` to access the function on the variable
- [ ] (Maybe?) Add to `catalog_controller` as a `search_field`

app/services/oregon_digital/triple_powered_service.rb
```diff
@@ -0,0 +1,131 @@
...
+      else
+        fetch_all_labels(uris).first
+      end
+    end
+
+    # METHOD: Functionality to fetch all labels and put it as 'label$uri'
+    def fetch_all_labels(uris)
+      labels = []
+      uris.each do |uri|
+        graph = fetch_from_store(uri)
```
CGillen, line 23: Ok, this one might be a difficult one to explain, so let me know if we need to jump on a call later.

I don't think we need the fetching part or `#predicate_labels` part of this service
Our controlled properties system is different from SA@OSU and already includes fetching. So we're able to do something like `work.subject` to give us an array of `OregonDigital::ControlledVocuabulary::Subject` objects, each with a `#fetch` method that will setup the object to have a readable label.
You should be able to get the `label$uri` setup with something like this:
```
subjects = object.subject
fetch_all_labels(subjects)

...

def fetch_all_labels(controlled_vocabs)
labels = []
controlled_vocabs.each do |cv|
  cv.fetch
  labels << "#{cv.rdf_label.first.to_s}$#{cv.rdf_uri.to_s}"
end
```

lamtu1: I will give this a try, if anything i will reach out. But thank you for the suggestion to fix this. I will get on to it.

lamtu1:
@wickr I got a chance to talked to Corey today and I wanted to follow up with you in regards of this `label$uri` feature implementation. I remember from the last work cycle for OD2, you wanted to meet up with me and Corey to discuss in regards of this ticket. And I was wondering if you still wanted to schedule a meet up with all of us to discuss any questions or concerns you have on this ticket?

app/services/oregon_digital/triple_powered_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module OregonDigital
+  # Triple Powered Service to fetch out data from uri and grouped as 'label$uri'
+  class TriplePoweredService
```
CGillen, line 5: Maybe rename to `LabelFetchService` to better match OD2 naming scheme & `LabelParseService`
(Don't forget to rename file)

CGillen (changes requested):

app/services/oregon_digital/triple_powered_service.rb
lamtu1: Yes I will rename it to fit the scheme. I'm currently rebuilding the image trying to fix the `docker-compose test` but will update it once got everything to pass the test with the updated on there.

CGillen (approved):

PR #3009, opened 2023-12-14
Feature/route for content blocks

straleyb:
- Swapping back from tombstoned now allowed.

- Route added allowing the saving of content block information for tombstone.

- Visibility shifted to private when tombstoned: Should remove from all places where private works are hidden unless you have permissions.

### DIFF
--- a/app/jobs/oregon_digital/swap_visibility_job.rb
+++ b/app/jobs/oregon_digital/swap_visibility_job.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+# reindexes chunks of uris
+class OregonDigital::SwapVisibilityJob < ContentEventJob 
+  def perform(work, visibility)
+    work.visibility = visibility
+    work.save
+    Hyrax::VisibilityPropagator.for(source: work).propagate
+    OregonDigital::PermissionChangePropagationJob.perform_later(work)
+  end
+end
+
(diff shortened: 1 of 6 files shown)


app/services/hyrax/workflow/swap_visibility_to_private.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module Workflow
+    # Hooked into workflow actions. When a work is tombstoned it runs this module and swaps the visibility to private
+    module SwapVisibilityToPrivate
+      def self.call(target:, **)
+        model = ActiveFedora::Base.find(target.id)
```
CGillen, line 8: Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: target.id)

app/jobs/oregon_digital/swap_visibility_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Swaps the visibility of an object as a job
+class OregonDigital::SwapVisibilityJob < ContentEventJob
+  def perform(work, visibility)
+    work.visibility = visibility
+    work.save
```
CGillen, line 7: Hyrax.persister.save(resource: work)

CGillen (changes requested):
Just the new Valkyrie lookup and save methods

CGillen (approved):

PR #3016, opened 2023-12-19
Remove unnecessary text from under 'Public' visibility option

### LINKED ISSUE #3011: Remove unneeded text from Edit Form Visiblity box
### Descriptive summary

On the Edit Form, in the Visibility selection box, remove the display of text under the 'Public' option, starting with `Please note, making`

image

This isn't relevant for Oregon Digital, and removing it will free up vertical space in the right sidebar.

### Expected behavior

In the Visibility box, the text under 'Public' option only needs to read 'Make available to all.'

### Related work

### Accessibility Concerns

megwalt:
closes #3011 
Remove unnecessary text from under 'Public' visibility option when the radio button is selected.

Screenshot:
Public visibility option selected


CGillen (approved):
Looks good. Those integration tests have started becoming pretty flaky lately. I restarted them to see if it helps, otherwise I can work with you on figuring out what is going on with them if you would like

CGillen:
Ah, it's these:
https://github.com/OregonDigital/OD2/blob/master/spec/system/create_audio_spec.rb#L65
You can just change those to something else in the sidebar. like: `Make available to all.`

megwalt:
It seems that line was checking to see if the public option was selected. The 'Make available to all.' is always in the sidebar. Should I try to get an equivalent check to see if the public option has been selected in there instead?

megwalt:
> It seems that line was checking to see if the public option was selected. The 'Make available to all.' is always in the sidebar. Should I try to get an equivalent check to see if the public option has been selected in there instead?

On second thought, that line was probably checking to make sure that the text showed after the public option was selected, not that the option was successfully selected. I'll edit the text that's being checked.

CGillen:
> > It seems that line was checking to see if the public option was selected. The 'Make available to all.' is always in the sidebar. Should I try to get an equivalent check to see if the public option has been selected in there instead?
> 
> On second thought, that line was probably checking to make sure that the text showed after the public option was selected, not that the option was successfully selected. I'll edit the text that's being checked.

Yeah, I think you're right, it's just for checking to see if it pops up if public gets selected. I'm not too worried about that check

PR #3021, opened 2023-12-21
Disable deposit agreement check and relevant text.

### LINKED ISSUE #3012: Hide/disable Deposit Agreement check
### Descriptive summary

Requiring submitters to check the Deposit Agreement doesn't generally apply to us and comes more from a self-submit repository.

We should disable this check and remove it from the Requirements list.

image

We should also remove the checkbox and text for the Deposit Agreement.

image

From @carakey 
"we could have new folks agree to something like the deposit agreement as a condition of getting any kind of account with depositing privileges"

Since this feature comes from Hyrax, disabling it will be the least invasive, and if we decide we need it activated again, for everyone or for certain groups of users, we could adjust later.

### Expected behavior
Submitters are not required to check the Deposit Agreement box and it's not counted as a Required step.

### Related work

### Accessibility Concerns

megwalt:
closes #3012 

Screenshot:
Disabled deposit agreement check box and text

### DIFF
--- a/app/views/hyrax/base/_form_progress.html.erb
+++ b/app/views/hyrax/base/_form_progress.html.erb
@@ -8,9 +8,6 @@
           <% if Hyrax.config.work_requires_files? %>
             <li class="incomplete" id="required-files"><%= t('.required_files') %></li>
           <% end %>
-          <% if Flipflop.show_deposit_agreement? && Flipflop.active_deposit_agreement_acceptance? %>
-            <li class="incomplete" id="required-agreement"><%= t('.required_agreement') %></li>
-          <% end %>
         </ul>
       </fieldset>
     </div>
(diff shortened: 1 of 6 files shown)


app/views/hyrax/base/_form_progress.html.erb
```diff
@@ -8,9 +8,13 @@
           <% if Hyrax.config.work_requires_files? %>
             <li class="incomplete" id="required-files"><%= t('.required_files') %></li>
           <% end %>
-          <% if Flipflop.show_deposit_agreement? && Flipflop.active_deposit_agreement_acceptance? %>
+<%
+=begin%>
+ <% if Flipflop.show_deposit_agreement? && Flipflop.active_deposit_agreement_acceptance? %>
```
CGillen, line 13: These `Flipflop` lines indicate a setting from the dashboard:
image

Is disabling these two values enough to close this ticket as well?

CGillen (changes requested):

app/views/hyrax/base/_form_progress.html.erb
megwalt: I just checked, it looks like it would be enough to close the ticket.
Settings disabled
Menu without deposit agreement

CGillen: Awesome! can you change it on staging an move ticket to QA? When it passes you can go ahead and change it on prod too, no need to wait for next deploy round

CGillen:
Fixed with configuration

Issue #3024, opened 2024-01-26
Search Results for Complex Objects: 'Contains Items' is Maxing Out at 10


petersec:
### Descriptive summary

Search results for complex objects include 'Contains [some number] Items' in the metadata that displays on the search results page. The [some number] is supposed to indicate how many child items are available within the complex object. We've found that the system is maxing out the count of child objects at 10, such that complex objects that contain many more child items than 10 are not represented correctly. See for example this [search string](https://oregondigital.org/catalog?f%5Bnon_user_collections_ssim%5D%5B%5D=aerial-photos-mid-willamette&f%5Bresource_type_label_sim%5D%5B%5D=Complex+Object&locale=en&sort=title_ssort+asc).

### Expected behavior

The 'Contains Items' metadata field on a search results page should provide an accurate count of the number of child items contained within a given complex object.

### Accessibility Concerns

None

shieldsb:
QA fail.  Linked search above still shows a max of 10 items

lamtu1:
Hi @shieldsb can you list me a complex object you were looking at for me to check to see where it is failing? I double check on some on staging and seem to be working fine and wanted to make sure and continue with it.

shieldsb:
QA pass.

PR #3026, opened 2024-01-31
Add a way to count the correct child works

lamtu1:
`FEATURE:` Fix update to match the search result to amount of child works in a complex object.

### DIFF
--- a/app/views/catalog/_index_list_default.html.erb
+++ b/app/views/catalog/_index_list_default.html.erb
@@ -22,8 +22,9 @@
     </div>
   <% end %>
   <% if document.children? %>
+    <% @works = ActiveFedora::Base.find(document.id) %>
     <div class="metadata-row row">
-      Contains <%= document.children.count %> items
+      Contains <%= @works.child_works.count %> items
     </div>
   <% end %>
   <% if document.parents? %>


lamtu1:
fixes #3024

app/views/catalog/_index_list_default.html.erb
```diff
@@ -22,8 +22,9 @@
     </div>
   <% end %>
   <% if document.children? %>
+    <% @works = ActiveFedora::Base.find(document.id) %>
```
CGillen, line 25: We want to use ActiveFedora as little as possible, especially on pages related to the public user experience, as each query can take anywhere from 200ms to multiple seconds.
It might be possible to use `(document.member_ids - document.file_set_ids).count` to get the count of all members of the work, ignoring file sets

Extra note. with valkyrization, `ActiveFedora::Base.find` is replaced by `Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: id)`.
The most common replacements are:
```
ActiveFedora::Base.find(id)
  becomes
Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: id)

curation_concern.update_index
  becomes
Hyrax.index_adapter.save(resource: curation_concern)

curation_concern.save
  becomes
Hyrax.persister.save(resource: curation_concern)
```
And the general class hierarchy from parent to child is:
`Valkyrie::Resource`  =>  `Hyrax::Resource`  =>  `Hyrax::FileSet` (valkyrie version of fileset, instead of our `::FileSet`) || `Hyrax::Work`
`Hyrax::Work` => our work types

CGillen (changes requested):

Issue #1404, opened 2020-11-25
Rename some media facet values


CGillen:
### Descriptive summary
`Moving Image` and `Sound` aren't as descriptive or well understood as `Video` or `Audio`, respectively. We should use the more recognized terms.

### Expected behavior
`Moving Image` => `Video`
`Sound` => `Audio`

shieldsb:
QA fail.  Still shows moving image and sound.

wickr:
The applicable works on staging needed a reindex to use the new values, after doing that the facet values appear as expected.
![image](https://github.com/OregonDigital/OD2/assets/2293544/9228502d-5006-4030-8122-b3ad7e5f88c6)

The icon update in #3035 is useful though.
So probably QA Pass after the icon update.

shieldsb:
QA pass.

PR #3035, opened 2024-02-02
Update moving image/video icon

lamtu1:
`FIX:` Add in rake task to update any `moving image` or `sound` may persist in other old works.


lamtu1:
fixes #1404

wickr:
@lamtu1 The changes on your last PR were good, the relevant works on staging just needed a reindex. After that the facet values on staging look good. I left a comment on the ticket.
The icon update fix is still useful here.

The rake task is not needed at all for this type of update. We can run a solr search and find the works to update, then launch separate reindex jobs for each one.
Since the values in the work are not changing, there is no need to resave the whole work and write anything to Fedora. An `.update_index` would be enough to update Solr and be faster.
Also, loading up all the works in the system, and then checking if they need to be updated, in a single threaded way, would take probably several months to complete at this point. We would need to use separate jobs at a minimum, and try to minimize the first set of works to evaluate as much as possible.

lamtu1:
@wickr Thank you for letting me know in regards of that. I will just push a PR removing the rake task and it is good to know that we can just `.update_index` to Solr and that would be faster than going through the whole list of all works (and that would take ages).

PR #3037, opened 2024-02-07
Downcase all facet values for better a-z sorting

### LINKED ISSUE #1273: Blacklight facets sorted A-Z sorts regardless of uppercase/lowercase
### Descriptive summary
~If you chose `more` from a facet to get the all the facetable values. Sort by `A-Z Sort`. All values will be uppercase A-Z. At the end of uppercase will be lowercase A-Z. Facet values should be normalized to uppercase the first letter OR A-Z sort should have `A` and `a` next to each other.~

Edit:
Previously we attempted to sort uppercase before lowercase on the A-Z sort facet modal. This isn't actually possible, and needs to be reverted. A-Z Sort may be sorted with uppercase and lowercase intermingled, preferring correct alphabetical sort.

### Expected behavior
Facet A-Z sort is caseless sorted A-Z

CGillen:
Fixes #1273 
This will require a re-index, but it can be done before the code is deployed so the facets continue to work w/o delay.
The second phase of #3006 will also require a re-index which should also be possible before deploying the code all the way to prod (will need to deploy to workers first but not rails) so we should wait for that to merge.


CGillen:
Delay until next reindex

PR #3054, opened 2024-02-20
Add GrantEditGroupJob.

### LINKED ISSUE #3052: Add job to grant edit access to work and filesets to user groups
### Descriptive summary

We didn't get user groups set up and configured before migration, so any migrated works or since-created works won't have permissions set even if they are added to an Admin Set now. User groups make it easy to add edit permissions to works instead of for multiple people, but there are only pre-existing jobs for users. A job would help scale up the bulk assignment to handle tens of thousands of works. 

### Expected behavior

Bulk assignment of edit permissions to a user group on a work and fileset(s).

### Related work

### Accessibility Concerns

wickr:
Fixes #3052

### DIFF
--- a/app/jobs/oregon_digital/grant_edit_group_job.rb
+++ b/app/jobs/oregon_digital/grant_edit_group_job.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 #
 # Helper job to bulk grant edit permissions for a user group (role).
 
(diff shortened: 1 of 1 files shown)


CGillen (changes requested):
Can this be Valkyrie compatible?
```
r = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: pid) # Getting a valk resource
r.permissions_manager.edit_groups += [group_key] # Add group
r.permissions_manager.acl.save # Save permissions changes

file_set_ids(r).each do |file_set_id|
  file_set = Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: pid)
  file_set.permissions_manager.edit_groups += [group_key]
  file_set.permissions_manager.acl.save
end
```
This will automatically index the permissions to solr, no need to use `Hyrax.index_adapter.save(resource: r)`

PR #3056, opened 2024-02-20
Solrize CVs with consistant https scheme

### LINKED ISSUE #1342: Controlled vocab terms standardize the protocol on stored uris
### Descriptive summary
Controlled vocab terms should be permissive in what they accept and rigorous with what they store. Before a fetch, controlled vocab terms should accept either http:// or https:// in uris, but when they receive the response it should be stored back in a standard protocol.

This is to help with bulk editing later, when you have thousands of values in a column and trying to sort on that column sorts on protocol first.

@wickr I assume this should be configurable at the vocabulary level? So we can force all the uris from an individual authority to be the same.

### Expected behavior
URIs can come in as either http:// or https:// from the work form and bulk edit workflows, and the fetch request will gracefully handle communicating with the authority.

URIs are stored back in fedora and solr as either http:// or https://, configurable by the authority/vocab, regardless of input, request, or response.

CGillen:
Fixes #1342


CGillen:
Depends #2634

straleyb (approved):

PR #3057, opened 2024-02-21
add errors behavior to hyrax resource

### LINKED ISSUE #3053: adding errors to works
### Descriptive summary
Using ActiveModel::Errors is not working as expected. After discussing with gillenc, implementing a small error class that uses redis to stash errors for a given work. 

### Expected behavior
Class will support adding errors associated with a keyword to a given work, displaying all errors for a given work, and removing errors with a keyword for a given work.
### Related work

### Accessibility Concerns

lsat12357:
fixes #3053

### DIFF
--- a/app/services/oregon_digital/verify_work_service.rb
+++ b/app/services/oregon_digital/verify_work_service.rb
@@ -15,22 +15,22 @@ module OregonDigital
       @services.each do |service|
         verifier = service.new(service_args)
         verifier.verify.each do |k, v|
-          v.blank? ? work.errors.delete(k) : add_errors(k, v)
+          v.blank? ? work.remove_errors(k) : add_errors(k, v)
         end
       end
     end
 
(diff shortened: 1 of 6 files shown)


CGillen (approved):

app/services/oregon_digital/verify_work_service.rb
```diff
@@ -43,7 +43,7 @@ def resource_type
     end
 
     def work
-      @work ||= resource_type.constantize.find(@args[:pid])
+      @work ||= Hyrax.query_service.find_by_alternate_identifier(alternate_identifier: @args[:pid])
```
CGillen, line 46: A note for querying valkyrie:
We still have some issues (like indexing) that require an AF object. If this starts breaking, you can add `use_valkyrie: false` as an arg to return an AF object, but still keep us prepared to switch over once we have real valkyrie objects in place

PR #3063, opened 2024-02-23
Cleanup on Display labels & uris

lamtu1:
`UPDATE:` Continue to update more on the display usage of the new `label$uri`

`TASKS:` To-Do lists

- [x] Update the `data_source.html.erb` to not use the `zipped_values` method and use the new `label$uri` method
- [x] Update the `OregonDigital::MetadataDownload#metadata_row` to get the new metadata without doing much fetching and cleanup as well
- [x] Update the `OregonDigital::PresentsDataSources` as well
- [x] Update the `_attribute_row.html.erb` to not use the `zipped_value`
- [x] ~(Future Task?) TBD (Spot any new file that may use `fetch` or `zipped_values` to display)~

### DIFF
--- a/app/models/concerns/oregon_digital/metadata_download.rb
+++ b/app/models/concerns/oregon_digital/metadata_download.rb
@@ -17,7 +17,7 @@ module OregonDigital
         if (service = local_metadata_authority_service(label))
           from_local_authority(service, Array(values))
         elsif controlled_keys.include?(label.to_sym)
-          from_controlled_field(values)
+          from_controlled_field(label)
         elsif values.respond_to?(:map)
           values.map(&:to_s).join('|')
         else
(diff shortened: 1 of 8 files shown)


lamtu1:
fixes #2634

app/models/concerns/oregon_digital/metadata_download.rb
```diff
@@ -33,8 +33,15 @@ def from_local_authority(service, values)
       values.map { |prop| service.label prop }.join('|')
     end
 
-    def from_controlled_field(values)
-      values.map { |prop| controlled_property_to_csv_value(prop).to_s }.join('|')
+    def from_controlled_field(label)
+      # FETCH: Get the SolrDocument that store the label$uri
+      cv_values = SolrDocument.find(id)
```
straleyb, line 38: Same thing here

app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -1,7 +1,13 @@
+<%# FETCH: Fetch the label$uri by finding it in SolrDocument %>
+<% cv_label_uris = SolrDocument.find(presenter.id) %>
```
straleyb, line 2: Here too.

app/views/hyrax/base/_data_sources.html.erb
```diff
@@ -8,14 +8,15 @@
     </tr>
   </thead>
   <tbody>
+  <% cv_label_uris = SolrDocument.find(presenter.id) %>
```
straleyb, line 11: Same thing

straleyb:
My comments aren't deal breakers. Just something I noticed while reading to understand the code.

lamtu1:
> My comments aren't deal breakers. Just something I noticed while reading to understand the code.

Yeah I totally understand that is misleading on what they do. I will double check and rename them to make it better and easier to understand the variables name

straleyb (approved):

PR #3073, opened 2024-03-06
Disables dropdown and adds text to make data more clear

### LINKED ISSUE #3062: Edit Work > Sharing > Currently Shared With section is showing extra and duplicate values
### Descriptive summary

On Edit Work view, Sharing tab, Currently Shared With section, the list of users and groups is showing extra values, which can be confusing to read and edit.

image
https://staging.oregondigital.org/concern/images/t148fh186/edit?locale=en#share

If the access level is 'Edit' I don't think we need to show 'View/Download' as well, at least on a separate row.

And the multiple rows showing the same values can also lead to problems when changing one (which is rare), not sure how the form would handle different values submitted.

### Expected behavior

Currently Sharing With section shows correct and concise list of users and groups with access.

### Related work

### Accessibility Concerns

straleyb:
Fixes #3062

Me and Ryan met about this ticket. Its more work than worth to actually clean up the data. However, we added some data to help explain why there are duplicate name/rows and what the permission for those dups are. We also disabled the dropdown because the permissions don't make sense changing access level when we are expressing permissions differently.

Screenshot 2024-03-06 at 2 46 07 PM

### DIFF
--- a/app/views/hyrax/base/_currently_shared.html.erb
+++ b/app/views/hyrax/base/_currently_shared.html.erb
@@ -1,0 +1,47 @@
+<% permission_service = Hyrax::EditPermissionsService.build_service_object_from(form: f, ability: current_ability) %>
+
+<h2 class="h3 mt-4"><%= t('.currently_sharing') %></h2>
+
+<table class="table table-bordered">
+  <tr>
+    <th><%= t('.table_title_user') %></th>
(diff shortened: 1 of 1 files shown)


straleyb:
@wickr

wickr (approved):

PR #3108, opened 2024-04-24
Skeleton for content warnings

straleyb:
This doesnt fully fix 3094. It gets all the bones in place with temporary text in it to get ready for the setup of content warnings. It generates a new tab and fills it with the proper radio buttons and text box. It also gets the layout MOSTLY looking correct. Minor tweaks needed for SR stuff. 
Screenshot 2024-04-24 at 1 21 37 PM


app/helpers/hyrax/work_form_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+  module WorkFormHelper
+    def admin_set_options
+      return @admin_set_options.select_options if @admin_set_options
+
+      service = Hyrax::AdminSetService.new(controller)
+      Hyrax::AdminSetOptionsPresenter.new(service).select_options
+    end
+
+    # OVERRIDE FROM HYRAX. Adds in warning tab
+    def form_tabs_for(form:)
```
CGillen, line 14: @lamtu1 and I actually found you can shove this method into `app/helpers/hyrax_helper.rb` for less override.
https://github.com/osulp/Scholars-Archive/blob/main/app/helpers/hyrax_helper.rb#L26

CGillen (changes requested):
Move `#form_tabs_for` for less override

CGillen (approved):
Approved but not merging per deploy delay discussion

PR #3112, opened 2024-04-30
Use lowercased facets for advanced search

CGillen:
This is a live hotfix for production. Do not update branch or merge until `ReExtractTextWorker` jobs are finished and we are ready to deploy final changes for full text search rewrite


CGillen:
Not the solution we're going with

Issue #3029, opened 2024-02-02
Add Local Contexts Labels Controlled Field And Display Labels on Work Show Page


KevinJonesMeta:
### Descriptive summary

https://localcontexts.org/
Local Contexts labels allow OD institutions to identify works that contain biocultural heritage and/or traditional knowledge of indigenous communities and communicate to those communities opportunities for collaboration regarding those works. These labels contain both an image icon and brief text outlining the meaning of a label.

### Expected behavior
TK labels and BC labels from Local Contexts display in the metadata and designated locations on work show page; fields allow multiple labels and icons to display in relevant locations; labels are indexed and searchable; in ingest form label values are available in a dropdown list similar to creative commons license values; use label page URLs for field values (URIs don't exist as of 2/2/24); webpage with explanation of use of labels and source of labels exists in About/Dig Deeper section of Oregon Digital.

Known Collections That Will Use TK/BC Labels:

(UO) [The Healers Project Collection](https://oregondigital.org/collections/the-healers-project?locale=en)

Known Task List:

- [ ] Create metadata field for TK and BC labels, using https://opaquenamespace.org/ns/localContexts
- [ ] Create facet for Local Contexts field available for collections - defaults to off
- [ ] Create value list for ingest form and add field(s) to ingest form
- [ ] Display TK and BC label icons in upper section with viewer on work show page
- [ ] Display TK and BC label label title in metadata on work show page
- [ ] Explanatory webpage for Local Contexts labels (TK/BC labels) exists in Oregon Digital and linked as appropriate from collection pages
- [x] Add metadata fields to MAP (Metadata Team)

Local Contexts Label titles and URLs sheet: https://docs.google.com/spreadsheets/d/1_jJ22sTw5ys_prLOr0KF_6A-bMtOzqBkNWWuRlgfGUE/edit#gid=0
Style Guide: https://localcontexts.org/wp-content/uploads/2023/08/TK-and-BC-Labels-Usage-and-Style-Guide.pdf
About Page text: https://docs.google.com/document/d/1BniBKhXFGybxpDAJoaGm7-eZe2v9lO07/edit

### Related work

Link to related tickets or prior related work here.

### Accessibility Concerns

Add any information here to indicate any known or suspected accessibility issues for this ticket

wickr:
[local_contexts_icons.zip](https://github.com/OregonDigital/OD2/files/14300752/local_contexts_icons.zip)
[...]

PR #3126, opened 2024-05-10
Local Contexts on Collection

lamtu1:
`FEATURE:` Add local contexts field into the metadata of the collection page.

`TO-DO:`
- [x] Add in the `metadata` into the `collection form`
- [x] Check to see if the `LC` exist in `@collection`
- [x] Hook it up to the display to present the LC


lamtu1:
fixes #3029

CGillen (changes requested):
I'm so sorry. Could you make `local_contexts` singular to follow the rest of the metadata properties? => `local_context`

jsimic:
@CGillen Local Contexts is an organization, plural: [https://localcontexts.org/](https://localcontexts.org/)

app/views/hyrax/collections/show.html.erb
```diff
@@ -114,6 +114,14 @@
     <button type="submit" form="collection-search" class='btn btn-lg btn-primary collection-landing-button col-sm-10'><i class='glyphicon glyphicon-search'></i> Search</button>
   <% end %>
 
+  <!-- Create render display for the Local Contexts Icon -->
+  <% if !@collection.local_contexts.blank? || !@collection.local_contexts.empty? %>
```
CGillen, line 118: `#blank?` should act the same as `#empty?` across empty arrays, empty strings and doesn't trip up on `nil`. You shouldn't need the call to `#empty?` unless I missed something

app/views/oregon_digital/base/_local_contexts.html.erb
```diff
@@ -8,7 +8,9 @@
     <%# FETCH: Get the array of the local contexts from work/collection %>
     <% if location_name.eql?('collection') %>
       <h2 class='h6'>Local Contexts</h2>
-      <%# DISPLAY COLLECTION CODE WILL GO HERE ON ANOTHER PR %>
+      <% @local_contexts_labels = OregonDigital::LocalContextsService.new.all_labels(collection_lc.local_contexts) %>
+      <% @lc_labels = @local_contexts_labels %>
```
CGillen, line 12: Extra assignment?

CGillen (changes requested):

app/views/oregon_digital/base/_local_contexts.html.erb
lamtu1: I have to pass in an extra `assignment` as the row display at the bottom of the code utilize the `@lc_labels`. So to make sure that one section on display work for both `work & collection`, I have to do that extra assignment to make the display universal usage in both area of `work & collection`.

app/views/hyrax/collections/show.html.erb
lamtu1: Got it. I will go in and see through all the code and remove those extra `empty?` code, glad to know `blank?` does all of that, I thought it only handle `nil`

CGillen (approved):

PR #3130, opened 2024-05-21
Feature/refactor content warning

straleyb:


app/assets/javascripts/oregon_digital/custom-content-copy.js
```diff
@@ -0,0 +1,3 @@
+function copyContentWarningValue(event) {
+  document.getElementsByClassName("content_alert_custom_radio")[0].value = event.target.value;
```
CGillen, line 2: I think this can be generalized to something like
```
$(event.target).parent().find('input[type=radio]').val(event.target.value)
```

app/views/hyrax/base/_form_warning.html.erb
```diff
@@ -21,8 +21,7 @@
   <li class="radio">
     <label>
```
CGillen, line 22: I didn't catch the nested `label`s here. Not sure the solution, maybe individual `label`s with `for='<id>'`?

CGillen (changes requested):

PR #3136, opened 2024-05-22
Add in LC to other work models

lamtu1:
`FEATURE:` Add in update to the `works` and allow it to appear on the other works form.


lamtu1:
fixes #3029

CGillen (approved):

PR #3139, opened 2024-05-23
Toggles default based on value or selects other

straleyb:


straleyb:
This chooses the default value if no value exists or if the default text is detected. Otherwise it chooses other and prefills with the proper value.

app/views/hyrax/base/_form_warning.html.erb
```diff
@@ -14,14 +14,15 @@
 <ul>
   <li class="radio">
     <label>
-      <%= f.radio_button :content_alert, t('hyrax.content_warning.content_alert') %>
+      <% @checked = (f.object.content_alert.blank? || (f.object.content_alert == t('hyrax.content_warning.content_alert'))) %>
```
CGillen, line 17: Can you move the `@checked` definition out of the `<ul>` so it's clearer where it's going to be used?

app/views/hyrax/base/_form_warning.html.erb
```diff
@@ -14,14 +14,15 @@
...
+      <%= f.radio_button :content_alert, t('hyrax.content_warning.content_alert'), class='content_alert_box', checked: @checked ? 'checked' : '' %>
```
CGillen, line 18: You probably already tested this, but does the `checked` form option work w/ regular truthy values? or does it automatically just set it to whatever is passed in?
`checked: @checked` would be cleaner, or `checked: !!@checked` if need be, but might not be possible all together

CGillen: `<input ... checked="checked">` is the most correct, but not required (except for xHTML), `<input ... checked>` or w/ any value is sufficient

CGillen (changes requested):

app/views/hyrax/base/_form_warning.html.erb
straleyb: I havent tested to see if the simple form options can just push through the `checked` html. I can see, but I know that it passes the checked in as an option. I can give that a bit of attention to see if I can get that working and update here.

CGillen (approved):

PR #3141, opened 2024-05-24
Add hints to the Content Warning

lamtu1:
`FEATURE:` Add in `hints` to let the user know what the area they are selecting are for in content warning. Information are being fetched from MAP.


straleyb:
This should get rebased with the newest main branch stuff.

straleyb (approved):

straleyb:
@tu I think this needs `t('simple_form.hints.defaults.mask_content')` or did I miss something?

lamtu1:
@straleyb since there is only one `:mask_content` on the form. It automatically populated above `item show page` checkboxes. Since the `:content_alert` have two areas usage, it doesn't do automatically, so I have to manually shove it in with that `text` box (Shown in image above).

PR #3147, opened 2024-07-02
add parsable label to colls

### LINKED ISSUE #3145: collections edit page error
### Descriptive summary

When a user clicks on the edit button for a collection, the page fails to load, user gets the 500 page.
This doesn't appear to be a problem with the collection itself.
Actual error: ActionView::Template::Error (undefined method '[]' for nil:NilClass
More of the stack trace attached below.
Appears to be related to rendering controlled vocabs, and as basically all collections have at least institution, no collections are editable from the front end at the moment.

### Related work

### Accessibility Concerns

lsat12357:
fixes #3145

### DIFF
--- a/app/indexers/concerns/oregon_digital/parsable_label_behavior.rb
+++ b/app/indexers/concerns/oregon_digital/parsable_label_behavior.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+module OregonDigital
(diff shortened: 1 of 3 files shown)


lsat12357:
Needed to use the code for parsable labels from the generic indexer on collections as well; have pulled it out as a module and then included it in both places

CGillen (approved):

PR #3220, opened 2025-01-13
Fix warnings and errors from rubocop

CGillen:


CGillen:
Approved for override merge ✅

PR #3255, opened 2025-01-29
Show ordered children

### LINKED ISSUE #2945: Complex Objects: child works are out of order in 'Contained In This Object' interface at bottom of show page
### Descriptive summary

For complex objects, the child works rendering in the 'Contained In This Object' interface at the bottom of the show page are not in numerical order. See, for example, [this work](https://oregondigital.org/concern/generics/1544bq240).

opera-varia

### Expected behavior

The ordering of child works in the 'Contained In This Object' interface should match the ordering of child works as presented in Universal Viewer. Presumably, this will usually be numerical order beginning with child item 1. 

### Related work

#2786 addressed this problem for child works displaying in Universal Viewer.

### Accessibility Concerns

None

lsat12357:
fixes #2945 

Notes: staying within the search builder landscape, this is a moderately hacky way to get the documents in order: calculate which ids should be displayed on a given page and add a second search_builder that only requests those ids; then when the search builder returns the documents, put them in order (there are only 4 of them, so not so bad). The original search builder is retained to provide the total solr response, necessary for pagination.  WRT to pagination...the wonkiest bit is this detail in the ordered_children_of_work_search_builder: in order to get results back, the query must show that this is a first page request, otherwise the logic says that you are building page 2 but there are only 4 items, which would have been seen on page one, and therefore there are no more documents to view.

### DIFF
--- a/app/controllers/concerns/oregon_digital/work_relation_pagination_behavior.rb
+++ b/app/controllers/concerns/oregon_digital/work_relation_pagination_behavior.rb
@@ -25,8 +25,14 @@ module OregonDigital
 
     # Pass work to relationship search builders
     def search_builder
-      work ||= ::SolrDocument.find(params['id'])
       search_builder_class == Hyrax::WorkSearchBuilder ? search_builder_class.new(self) : search_builder_class.new(self, work: work)
+    end
+
+    def work
+      @work ||= begin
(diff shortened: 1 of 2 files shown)


lsat12357:
Screenshot at 2025-01-29 13-44-22

wickr (approved):

Issue #3245, opened 2025-01-22
UX Homepage Featured Collections


azle:
### Descriptive summary

When using the screen reader the Featured Collections is reading the image thumbnails alt tag before reading the collection title. It is confusing to a non visual user because the alt tags doesnt reference the collection title.

### Expected behavior

As a user, I expect the title to be read first before the image alt tags to know that these thumbnails belong to this collection.

### Accessibility Concerns

Issues with the screen reader reading the thumbnail images before stating the collection title.

lamtu1:
There is spacing image issue on `feature_collection` which I will get a fix up to make evenly distributed

lamtu1:
QA Pass

PR #3324, opened 2025-04-11
Add in title first then thumbnail

lamtu1:
`UPDATE:` Move the title of the `feature-collection` at the top so the screen reader will read it first, but still keep the order intact on the layout.


lamtu1:
fixes #3245

PR #3348, opened 2025-04-25
use triplestore

### LINKED ISSUE #2461: subject controlled vocabulary logic for handling itis
### Descriptive summary
We have special handling for itis uris that were added to scientific, but not subject.

### Related work

### Accessibility Concerns

lsat12357:
fixes #2461


lsat12357:
replaces the fetch method in subject.rb with the one used in scientific.rb; uses super unless class is itis.

wickr (approved):

PR #3418, opened 2025-09-24
Keep query params after redirecting user from cloudflare challenge

### LINKED ISSUE #3417: Searches from the homepage are not persisted for users who haven’t first been Cloudflare authenticated
### Descriptive summary

If a user hasn't been authenticated by Cloudflare and they conduct a search from the Oregon Digital homepage (banner box or center box), their search terms are not persisting to the search results page. 

Instead, once they hit Enter to run their search, Cloudflare authenticates them and, after having done so, they arrive at the search results page but are met with the results of an empty search. Their search terms don't persist anywhere on the page and in order to correctly run their search, they are forced to remember and re-enter their original search terms in the banner box or elsewhere. 

This issue goes away once a user has been authenticated by Cloudflare, for however long that authentication lasts. But it returns once the user's authentication has expired.  

### Expected behavior

The user should be able to run their search from the homepage, receive authentication from Cloudflare, and then be passed through to the search results page with their search terms having persisted. The search results page should return the results of the search that they conducted rather then presenting the results of an empty search. Their search terms should also remain visible in the breadcrumbs on the search results page.

### Accessibility Concerns

None.

CGillen:
Fixes #3417

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.7-alpine3.15 as bundler
+FROM ruby:2.7-alpine3.15 AS bundler
 
 # Necessary for bundler to operate properly
 ENV LANG C.UTF-8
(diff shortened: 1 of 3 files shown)


CGillen:
Yup, I can do that

straleyb (approved):

Issue #3453, opened 2025-10-21
Add Administrative Metadata section to bottom of work show page


wickr:
### Descriptive summary

There's some administrative metadata that the system knows but is sometimes buried or tough to find without going to the rails console.

For admins, add a section at the very bottom of the work show page that includes the following:

* Last Indexed Date: [from solr doc, timestamp field]
* Admin Set: Title and ID [link to Admin Set dashboard works view]
* Depositor: [link to search results for depositor?]
* [if applicable] Bulkrax Import ID:  [link back to Importer page, not the Entry directly, just to save another ID lookup]

### Expected behavior

Example work:
https://oregondigital.org/concern/images/6w924d53d

* Last Indexed Date: 2025-10-21T22:52:45.278Z
* Admin Set: osu-scarc-admin
* Depositor: joneske3@oregonstate.edu
* Bulkrax Importer ID: 3649-21

### Related work

Somewhat related to #3399 

### Accessibility Concerns

Add any information here to indicate any known or suspected accessibility issues for this ticket

lamtu1:
@wickr I was wondering if you could explain more detail on how the linking of each work? Like for the `admin set, depositor, and importer ID`?

And also, below I have a markup of how it looks like was wondering if this is good and if there are any detail or thing you wanted to change in regards of which data need to display or change layout of the area.

<img width="1266" height="426" alt="Image" src="https://github.com/user-attachments/assets/cd2c9542-9f1f-4f6b-9317-7caa5d8af2a4" />


Below here is one with importer ID:

<img width="1266" height="515" alt="Image" src="https://github.com/user-attachments/assets/4f4d7be0-3891-48c7-bc9b-98b861187718" />

wickr:
Also confirming for the Bulkrax Importer, it'd be the entire ID (like `3649-21`), but I think just link to the Importer page.

PR #3481, opened 2025-11-12
Add in Metadata for admin

lamtu1:
`FEATURE:` Create a section for admin to view the works metadata.


lamtu1:
fixes #3453

lamtu1:
Screenshot 2025-11-12 at 12 04 51 PM

wickr:
It's been a while since I looked at the page display code, but is there a way to avoid yet another SolrDocument call?  Since basically that's what powers the show page in the first place, I thought all of this info might already be included in some form and wouldn't need another query, that would slow things down even further.

straleyb (approved):

Issue #3535, opened 2026-02-17
Accessible Copy Request Form Only Has Heading On Page (Staging)


KevinJonesMeta:
### Descriptive summary

The Accessible Copy Request link on the About page opens a subpage that has a heading 'Accessible Copy Request' but the form is missing.

### Expected behavior
The Accessible Copy Request form is on Accessible Copy Request page.

### Accessibility Concerns

Yes. Needed to receive accessibility remediation requests.

PR #3536, opened 2026-02-18
[v4] Update by adding in Accessibility Copy Form to v4

lamtu1:
`UPDATE:` Update the v4 to have the accessibility copy form.


lamtu1:
fixes #3535

Issue #3637, opened 2026-05-26
Extract Embedded Alt Text To Alt Text Metadata Field On Fileset and Display on Work Showpage


KevinJonesMeta:
### Descriptive summary

Alt text values (primarily on image files) will be embedded on files to support WCAG 2.1 compliance. Oregon Digital needs to extract that embedded metadata and populate an alt text field on fileset to display in file detail section of Works and Files on showpage. this value will also get added to IIIF manifest to travel with the work if the manifest is used elsewhere (see related ticket).

Metadeities and other submitters will populate both the IPTC Alt Text field and the ImageDescription field on files to ensure technical compliance and functional utility of the alt text value. ImageDescription appears to be an XMP field whose value is duplicated in the IPTC Description field. See p120_3375_bd.tif for an example. 

Metadeities has created a predicate to hold alt text values on the fileset:

http://opaquenamespace.org/ns/altText

Example files for reference (shared with @CGillen in box):

p120_3375_bd.tif - This file has a value in the IPTC Alt Text field and XMP ImageDescription fields on the file. The ImageDescription field value also populates the IPTC Description field on the file. the ImageDescription tends to be more readily accessible to image display apps but it is not an Alt Text field per se.

p120_3376.tif - This file has a value in the IPTC Alt Text field. Though using the correct field according to the IPTC standard, it does not display in many image display apps.

### Expected behavior

A file that is imported via Bulkrax or the UI form that contains embedded alt text metadata in the ImageDescription field has that data extracted and it populates the alt text field on the fileset. That fileset field displays on the showpage in the same area as the accessibilityFeature values for each fileset.

### Related work

#3636 

### Accessibility Concerns

Add any information here to indicate any known or suspected accessibility issues for this ticket

lamtu1:
@KevinJonesMeta I was wondering if you could link up to the work with `p120_3375_bd.tif ` and `p120_3376.tif` so I can see and understand the metadata and the content of it while I am trying to plan out for it? Thank You!

@briesenberg07 Thank you for the new predicate and I was also wondering if the MAP will be updated for this predicate in our drive by any chance? Thank you

KevinJonesMeta:
@lamtu1 I shared the files with you in Box. Is there something else that would help?

lamtu1:
> [@lamtu1](https://github.com/lamtu1) I shared the files with you in Box. Is there something else that would help?

[...]

PR #3657, opened 2026-06-08
Update bulkrax to get alt_text to work for FileSet

lamtu1:
`UPDATE:` Create an update to remove some function of `alt_text` in Bulkrax due to it being at work level not FileSet level.


lamtu1:
fixes #3637

CGillen (approved):

Issue #1560, opened 2021-03-18
[Notifications] Failed fetch jobs


CGillen:
### Descriptive summary
This run of tickets is based on the discussion in #230.
- Failed Fetch Jobs
  - An email should be sent to the depositor of the work
  - Explore the option of digesting all of these to the head of metadeities
  - Emails should come in batches at the beginning or end of the day

When a work has controlled metadata and the URI fails to resolve, a notification is sent to depositor of the work. We want those emails to come in batches at the beginning or end of the day. Head of metadeities should get a digest of these failed fetches as well; start with daily and see if we need to back off to weekly.

### Expected behavior
Failed metadata fetch errors notify depositor and digest to head of metadeities

### Related work
#230 

### Accessibility Concerns
None

lamtu1:
QA: Same as ticket #1562 (Need dev to QA this)

wickr:
QA Fail.
#1562 is passing, but at the end of that email output, it looks like the failed fetch part starts and then aborts:
```
rake aborted!
Errno::ENOENT: No such file or directory @ dir_initialize - ./tmp/failed_fetch/
/data/lib/oregon_digital/failed_fetch_notification.rb:61:in `open'
/data/lib/oregon_digital/failed_fetch_notification.rb:61:in `foreach'
/data/lib/oregon_digital/failed_fetch_notification.rb:61:in `block in create_zip_file'
/usr/local/bundle/gems/rubyzip-2.3.2/lib/zip/file.rb:125:in `open'
/data/lib/oregon_digital/failed_fetch_notification.rb:59:in `create_zip_file'
/data/lib/tasks/oregon_digital/daily_email.rake:27:in `block (2 levels) in <top (required)>'
/usr/local/bundle/gems/rake-13.0.6/exe/rake:27:in `<top (required)>'
Tasks: TOP => oregon_digital:daily_email
(See full trace by running task with --trace)
```

wickr:
Closed by #3341

lsat12357:
FYI to help with QA I imported a new item with a creator that would fail to fetch a label: https://staging.oregondigital.org/concern/images/cf95jb59d?locale=en#data_sources

wickr:
QA Fail.
Unless I'm missing something, still not seeing any failed fetch emails in the task response. Not seeing anything created in `tmp` either.

lamtu1:
Will follow up with this ticket in the next workcycle and need to check in on the QA.

[...]

PR #3672, opened 2026-06-17
Store fail URI fetch into Triplestore

lamtu1:
`UDPATE:` Add the fail fetch to the `triplestore` rather than `workers` to do the work.

### DIFF
--- a/app/indexers/oregon_digital/deep_indexing_service.rb
+++ b/app/indexers/oregon_digital/deep_indexing_service.rb
@@ -57,11 +57,13 @@ module OregonDigital
 
     # override Hyrax fetch_value to get rid of misleading logging
     def fetch_value(value)
-      Thread.current[:work_id] = @object.id
       value.fetch(headers: { 'Accept' => default_accept_header })
     rescue IOError, SocketError, TriplestoreAdapter::TriplestoreException => e
       # IOError could result from a 500 error on the remote server
       # SocketError results if there is no server to connect to
+      # Add in the fail fetch attempt to store the info what is being fail
+      OregonDigital::Triplestore.create_and_check_directory
+      OregonDigital::Triplestore.store_failed_fetch(@object.id, [{ uri: value.rdf_subject.to_s, error: e.message }])
       Rails.logger.error "Unable to fetch #{value.rdf_subject}.\n#{e.message}"
     end
 
--- a/app/indexers/oregon_digital/deep_indexing_service.rb
+++ b/app/indexers/oregon_digital/deep_indexing_service.rb
@@ -57,6 +57,7 @@ module OregonDigital
 
     # override Hyrax fetch_value to get rid of misleading logging
     def fetch_value(value)
+      Thread.current[:work_id] = @object.id
       value.fetch(headers: { 'Accept' => default_accept_header })
     rescue IOError, SocketError, TriplestoreAdapter::TriplestoreException => e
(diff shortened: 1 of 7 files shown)


lamtu1:
fixes #1560

app/indexers/oregon_digital/deep_indexing_service.rb
```diff
@@ -57,6 +57,7 @@ def fetch_external
 
     # override Hyrax fetch_value to get rid of misleading logging
     def fetch_value(value)
+      Thread.current[:work_id] = @object.id
```
CGillen: Since we're running upwards of 15+ pods/instances, I don't think the `Thread` library will work for data sharing

app/workers/fetch_graph_worker.rb
```diff
@@ -50,34 +40,4 @@ def fetch_failed_graph(val)
...
-      end
-      f.puts("Time Recorded: #{Time.now.strftime('%m-%d-%Y %I:%M %p')}")
-    end
-  end
-  # rubocop:enable Metrics/AbcSize
-  # rubocop:enable Metrics/MethodLength
-
-  # METHOD: Create and check exist of folder
-  def create_and_check_directory
-    dir_path = Rails.root.join('tmp', 'failed_fetch').to_s
```
CGillen, line 80: I don't think `tmp/failed_fetch` is a shared directory either. `tmp/shared` is shared across all pods, so `tmp/shared/failed_fetch` should work if that is why you're moving away from this approach

CGillen: Alternatively, the less safe (safety in this context isn't really important imo) but faster and clean way would be to dump these values into redis, though that is probably much more annoying

CGillen (changes requested):

app/workers/fetch_graph_worker.rb
lamtu1: Okay I will look into dumping the content in the shared folder

app/indexers/oregon_digital/deep_indexing_service.rb
lamtu1: Got it, I will look into a way to access the `id` to get it into the triplestore in a better way

lib/oregon_digital/failed_fetch_notification.rb
```diff
@@ -19,11 +20,18 @@ def add_list_of_users
       pid_arr = fetch_pids
 
       # LOOP: Go through Solr Doc to find depositor to user_map
-      solr_docs = SolrDocument.find(pid_arr)
+      solr_docs = pid_arr.filter_map do |pid|
```
CGillen: What was going wrong here? We specifically had to modify `SolrDocument#find` to accept an array of pids and return all matching docs for those IDs in a single query. If that's not working, that's a bigger fix that needs answered

Otherwise, if you absolutely need to know which docs didn't come back, you could do different set operations (union, diff, etc) between the returned IDs and the `pid_arr` to figure out whatever you might need

CGillen (changes requested):

lib/oregon_digital/failed_fetch_notification.rb
lamtu1: I think I was going with handling of just in case if a solr_doc couldn't be find and breaking the code, but since we handle that with the `.find` function, I can just reverted then.

CGillen (approved):

Issue #3671, opened 2026-06-17
Change Privacy Setting On Files Behavior


KevinJonesMeta:
### Descriptive summary

When a file is set to private on production the following behavior is happening.
1. Derivatives are present in UV Content panel
2. Fileset title (Title in Works and Files list) is set to `File`
3. Download button in Works and Files list for that file is removed.

Example: https://oregondigital.org/concern/documents/fx719t647

### Expected behavior

1. File pages/images do not appear in viewer
2. Fileset does not appear in Works and Files list (confirm?)
3. Download button in Works and Files list for that file is removed (if the file is in list) (current behavior) (Confirm?)

### Related work

Link to related tickets or prior related work here.

### Accessibility Concerns

Add any information here to indicate any known or suspected accessibility issues for this ticket

KevinJonesMeta:
@jsimic Can you comment on this ticket about what behavior you want to see when file is set to `private` for non-admin users and tag Corey?

KevinJonesMeta:
@CGillen Copied from jsimic in Slack for tracking purposes: I don’t have any memory of a decision on private files beyond collections and works. Did a quick search in the Drive, and I see nothing regarding files. If you’re asking for an opinion, I agree with Kevin that they shouldn’t show up in the viewer at all to a non-owner, nor should they be able to download. But I wouldn’t expect the files to be visible in Works and Files to the non-owner either. Only the owner should see them in the viewer, W&F, and have a download button.

KevinJonesMeta:
@lamtu1 Is this deployed to Staging for QA?

lamtu1:
@KevinJonesMeta Yes it should be on staging

KevinJonesMeta:
@lamtu1 should I expect the new behavior to happen with existing works or only new works?

lamtu1:
@KevinJonesMeta It should work on existing or new work created, so any is fine

KevinJonesMeta:
QA Fail. @lamtu1 

When not the owner/admin of work or not logged in: (Example: https://staging.oregondigital.org/concern/documents/vx021f16r)

- [x] File not available in Works and Files when set to Private
- [ ] File pages/images are not available and visible in Universal Viewer when fileset is set to Private

<img width="1505" height="848" alt="Image" src="https://github.com/user-attachments/assets/6f5693f0-d886-45f4-a212-d0f78743692b" />

When logged in as Admin (Should be true for owner of work as well): 

- [x] File available in Works and Files when set to private
[...]

PR #3677, opened 2026-06-25
Add in a role-based on cache for UV

lamtu1:
`UPDATE:` Add in a role-based into the cache system of `IIIF` to tell it which display will present based on the user status role.


lamtu1:
fixes #3671

CGillen (approved):

PR #3678, opened 2026-06-25
Display oembed viewer above uv if both available in children

### LINKED ISSUE #3608: oEmbed: Kaltura player not displaying in Universal Viewer for Complex Objects
### Descriptive summary

oEmbeds of the Kaltura media player are not displaying in Universal Viewer when included as parts of Complex Objects. For OSU, these objects have thus far tended to be oral history interviews that include the transcript of the interview as one child item and the interview recording (audio or video) as the other child. 

When a user lands on the show page for complex objects of this sort, they first see the Thumbnail view in the left UV menu, which displays thumbnails of the first several page of the transcript. If they click on Index view within that same menu, they see a link option for the interview recording, but if they select that option, the player just spins within UV and doesn't load. Here are a couple examples:

- [Clara Pratt oral history interview (video)](https://oregondigital.org/concern/generics/n009w330q)
- [C.V. Ruzek oral history interview (audio)](https://oregondigital.org/concern/generics/3484zh982)

In both of these examples, the interview recording does render properly as an oEmbed object unto itself (aka, when not part of a complex object). Here are links to those two objects:

- [Clara Pratt interview recording (video)](https://oregondigital.org/concern/videos/pr76f4480)
- [C.V. Ruzek Oral History Interview Recording (audio)](https://oregondigital.org/concern/audios/q524jq07v)

### Expected behavior

When the user navigates to the Index menu view within Universal Viewer and then clicks on the link for the recording, the Kaltura player should load, play and be usable within UV. The user should also have the ability to exit out of the player, return to the Index menu view and interact with other components of the complex object, in this case the transcript.  

### Related work

[Improve UV Contents pane functionality for complex objects containing audio or video  #2808](https://github.com/OregonDigital/OD2/issues/2808) [An older ticket that documents similar issues with UO complex objects, presumably using Panopto rather than Kaltura as the media host]

### Accessibility Concerns

None.

CGillen:
Fixes #3608 
@petersec @jsimic It's a bit of a rough fix but UV isn't super stoked about handling iframes, so the video can't go into UV.
image

### DIFF
--- a/app/actors/oregon_digital/actors/create_with_oembed_url_actor.rb
+++ b/app/actors/oregon_digital/actors/create_with_oembed_url_actor.rb
@@ -42,7 +42,6 @@ module OregonDigital
       # rubocop:disable Metrics/MethodLength
       def create_file_from_url(env, uri)
         oembed_url = URI.decode_www_form_component(uri.to_s)
-        use_valkyrie = false
         fs_class = env.curation_concern.is_a?(Valkyrie::Resource) ? Hyrax::FileSet : ::FileSet
 
         file_set = fs_class.new(oembed_url: oembed_url) do |fs|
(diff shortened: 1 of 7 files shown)


petersec:
Thanks very much @CGillen. From the screenshot this looks like it will get the job done. 

Is this something that we'll want to test in Staging? I don't think the needed objects are in Staging right now, but @KevinJonesMeta might be able to get them in tomorrow or I could add them next week. (I've been out most of today and am out all day tomorrow.) This is also definitely something that can wait until the next work cycle if we need to fine tune.

straleyb (approved):

Issue #3706, opened 2026-08-05
Large collection downloads produce corrupt zip files


CGillen:
### Descriptive summary
@KevinJonesMeta found that collection downloads are not working correctly for at least https://staging.oregondigital.org/collections/daily-barometer, likely large collection in general though.

https://github.com/OregonDigital/OD2/issues/3690#issuecomment-5185066785

> For The Daily Barometer Collection https://staging.oregondigital.org/collections/daily-barometer:
    - Download button triggered download of zip folder
    -  Download finished; size 1.8 GB
    - When attempting to extract folder, zip was treated as empty (see image below)

### Expected behavior

1. Collection download button streams a zip file to the user
2. Finished zip file can be extracted
3. Contents of extraction is matching metadata, thumbnail images or other files, all open-able.

### Accessibility Concerns
N/A

CGillen:
Fixed by #3708

KevinJonesMeta:
QA Fail. I got the message below when I tried to unpack the new zip from the collection linked above on Staging.

<img width="957" height="664" alt="Image" src="https://github.com/user-attachments/assets/7e6dd5b3-5814-45d0-ae55-809b98c6caa7" />

CGillen:
Not fixed, reopening

straleyb:
This is actually a compound of multiple problems. When there is messed up data in the download, it will break it. The messed up data is a data cleanliness problem. 

A second was because the streaming service was trying to write derivatives that didnt exist. This is now fixed by #3718 

The third is LARGE file downloads dont work because of timeouts.

lamtu1:
Through research, looks like we are pulling all the `ActiveFedora` object into streaming and going through each one-by-one. The thing is we also needed Fedora for all the metadata stuff being created for the CSV file. With that, there was another method for stream in `ZipKit` called `write_stored_file` to help speed up some specific file for download but it is not a good method after testing because we need to know the size of the file before it stream so good safe and fast reliable one we use right now is the only way with `write_deflated_file` method.

briesenberg07:
per standup 2026-08-14, there are two distinct issues:

- 1️⃣  **data corruption**
- 2️⃣  **streaming**

QA details for only 1️⃣ **data corruption** -- I'm experiencing issues opening unzipped .tif files 

"Download Collection" from https://staging.oregondigital.org/collections/daily-barometer
- download w/ Chrome Version 151.0.7922.137 (Official Build) (x86_64)
[...]

PR #3708, opened 2026-08-06
Replace zip_tricks with zip_kit

CGillen:
This will update our zip library to the successor of zip_tricks.
Adds a little extra error checking to prevent zips from crashing downloads
Better support PDFs


straleyb (approved):

CGillen:
Fixed #3706

Issue #3690, opened 2026-07-14
Hyrax v5 Testing on Staging


shieldsb:
The primary concern is data integrity, we don't want adds/edits/saves/review/imports/exports overwriting or deleting good data. We also want to know if search/indexing/display/etc still works.

Some testing ideas (non-exhaustive):

- [x] Create a work
- [x] Edit a work
- [x] Version a fileset
- [x] Review a work
- [x] Create a collection
- [x] Edit a collection
- [x] Add work(s) to a collection
- [x] Embargo/Lease stuff
- [x] Examine all pages for load & content (even if it looks funky)
- [x] Facet a search
- [x] Use a facet from the facet modal
- [x] Try paging through pagers
- [x] Advanced search
- [x] Bulkrax importer
- [x] Bulkrax exporter
- [x] Open all the modals
- [x] Spotlight work/researcher - N/A
- [x] Other things you do on a frequent basis?
- [x] Expected behavior
- [x] The site is reliable and doesn't destroy our data.

Accessibility Concerns
Most of our edits are still there, but if you see anything flagrant, flag it.

petersec:
**Observations from one test object**

I created a test object via the form, and then edited and reviewed it - https://staging.oregondigital.org/concern/images/kh04dp82v

I received an email when I completed the creation of the object ("Mailboxer new message: Deposit needs review") and another when it went live ("Mailboxer new message: Deposit has been approved"). I'm guessing this is what is expected but wonder if we'll want to retain it.

Global search, Advanced Search (using the `Creator` field) and Search Within all worked as expected. 

All Download options worked as expected.

Most faceting appears to work as expected, though see below for issues related to `Work Type` and `Date`.

**Issues that I found:**

I was not able to assign the `Collection` that I wanted. I wanted to add it to `osu-historical-images` , which definitely exists in Staging, but the dropdown form on the Relationships page wouldn't find it. I didn't see any other way to add a collection via the form.

I also initially didn't think to set the item visibility to `Public` before reviewing. Once I realized I needed to take that extra step, I selected `Public` under Visibility and was taken to a page asking if I wanted to "Apply changes to contents?" I selected `Yes Please` and was met with a `404 Not Found` error. The requested action did go through, but as the user it was up to me to find the object in Staging and verify that the visibility had been changed. I've repeated this process twice - going from Public to Private and back to Public - and have encountered the same 404 error each time.

[...]

PR #3709, opened 2026-08-06
Fix indexer lookup for Valkyria => PCDM hack

CGillen:
Fixes https://github.com/OregonDigital/OD2/issues/3690#issuecomment-5207650553


straleyb:
Merging still because of Flakey Tests

PR #3721, opened 2026-08-14
Make sure Collection download doesn't pull from Fedora before stream starts

CGillen:
This should fix timeouts on collection downloads: https://github.com/OregonDigital/OD2/issues/3706#issuecomment-5285878348


straleyb (approved):

CGillen:
I'm forcing merge because the final broken tests are the flaky ones and I'm tired of waiting

